PR #3, opened 2020-01-19
Merge indirect and direct free kick into (direct) free kick

g3force:
From the Rule Change proposal:

> There are only a few situations, where an indirect free kick is effective:
> * Throw in -> large distance to goal, unlikely to score a goal directly anyway
> * Attacker in defense area -> Even larger distance to goal
> * Dribbling, ball speeding -> Rules will be changed such that the game is not stopped anymore
> 
> But it complicates things for:
> * The [human|auto] ref: Has to decide what kind of freekick has to be issued
> * The autoRef: Has to track indirect goals
> * GameController (GC): Majority between autoRefs more complicated
> * Teams: Have to distinguish free kicks (or they simply ignore it already)
> 
> We thus propose to merge both freekicks into direct freekick.


tobiasheineken:
This list does not contain all possible reasons for Indirect free kicks. As mentioned in #17, an indirect free kick is given for lack of progress - which can happen very close to the defense area if the keeper holds the ball.

The rules do allow for an defender to be placed on the direct line between the ball and the goal, but teams _can_ ignore this as another **attacking** robot needs to touch the ball for a valid goal. Which makes the position central in front of the defense area less scary than a direct free kick, as a stray shot will never be a valid goal.

@g3force mentioned the option to change to replace all **direct** kicks to **indirect** instead. While I think this is more reasonable, offering for more interesting strategies for defending, and as a direct result, for attacking, I also think this is no good solution either:
Curved shots, which have been mentioned in multiple (E)TDPs in the last years, always claim to be especially useful for corner-kicks and other static situations. This is mostly due to the fact that the kicking device can not be rotated quickly (yet), and the different arc can surprise the defenders that would normally block the direct line.
While I still haven't seen a very impressive or dangerous curved shot yet, the idea is interesting and perfecting it should be incentivized.  Removing situations where a slowly moving kicking device is useful does the opposite, as the barrier until this technology can be used in game is raised considerably.

All in all, I think we're better off without this PR. Different situations require different reactions. A stationary ball close to the goalkeeper where no attacker may enter is less dangerous than other situations that result in a (direct) free kick, like multiple defenders partially, where a fast moving ball might be deflected to miss the goal. Consequently, if the defending team fails to obey the rules the attacking team should be compensated differently.

rhololkeolke:
@tobiasheineken #17 removes the free kick in case of lack of progress and just uses a force start. 

I think merging these would simplify the game a lot by removing an entire state and removing the need to decide/remember which kick type is appropriate given what just happened in the game.

If the concern is that it is too easy to score with a direct free kick, I would rather change all free kicks to indirect free kicks than to keep the split around. Alternatively, if the concern about direct free kicks is only because of a few edge cases (like lack of progress in the defense zone) I would rather adjust where the kicks are taken from as the ball placement rules are already very situational, so we wouldn't be making things a lot more complicated by tweaking those.

andre-ryll:
I don't see the point to change all freekicks to indirect and would prefer the propsed solution of only having direct kicks. If there is a freekick in front of the defense area just place your robots in the direct shoot line, simple and effective. Curved shots are still and advantage here as they do not follow the direct line.

Furthermore, having all free kicks being indirects would complicate the game for viewers, refs, autorefs, etc. They all need to check if another robot touched the ball before a goal. That's hard for a referee and confusing for viewers. (Offside in regular soccer is a frequent source of irritation).

PR #4, opened 2020-01-19
Remove the double touch rule

g3force:
From the Rule Change proposal:

> After free kicks and kickoffs, robots are not allowed to touch the ball a second time until another robot has touched the ball.
> On the road to reducing stoppages, we propose to remove this constraint.
> Double touch in the SSL often happens due to imprecision, not by intention. When the ball is moved (by 0.05m), it is in play anyway, so all robots can touch it.
> We want to encourage teams to work on dribbling balls (dribbling in the sense of small kicks, the 1m dribbling rule still applies) and the double touch rule rather counteracts this.

### DIFF
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -5,9 +5,6 @@ When the match is <<Stopping The Game, stopped>>, the ball is considered *out of
 
 When the match is <<Resuming The Game, resumed>>, the ball is considered *in play* until the next stoppage occurs. The match is resumed when <<Force Start, force start>> has been issued or the ball moved at least 0.05 meters following a <<Kick-Off, kick-off>>, <<Direct Free Kick, direct free kick>>, <<Indirect Free Kick, indirect free kick>> or <<Penalty Kick, penalty kick>>.
 
-NOTE: see <<Double Touch, double touch>> for the rationale of the 0.05 meter distance
-
-
 === Ball Manipulation
 Shooting and <<Dribbling Device, dribbling>> is considered as manipulating the ball, the ball accidentally bouncing off the hull is not.
 
@@ -61,7 +58,6 @@ Chapter <<Offenses>>, section <<Minor Offenses>>:
 
 | <<Aimless Kick [small]#(_division B only_)#, Aimless Kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
 | <<Lack Of Progress>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
-| <<Double Touch>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
 | <<Attacker In Defense Area>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
 | <<Attacker Touches Robot In Opponent Defense Area>> skipped | <<Ball In And Out Of Play, ball in play>> | no command | icon:check[role="green"] (<<Advantage Rule>>)
 | <<Excessive Dribbling>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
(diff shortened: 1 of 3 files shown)


g3force:
Regarding the emailed comment from RRT:

> We agree with the removal of double touch, since this will encourage the teams to experiment with dribbling by small kicks. We do however wish that the double touch foul remains valid during penalties (if they are not removed) since this allows for unfair gameplay (and is also not allowed in real soccer).

Note that the ball will still be in play if it moved for more than 5cm and that it is not allowed to shoot a goal from within the defense area. A team could in theory pull the ball backwards/sidewards, but I currently do not see where a team would get an advantage from this.

As the penalty kick will be removed anyway, this is not relevant anymore. I just keep this here for reference, just in case.

tobiasheineken:
I still claim that this change is unnecessary.
Teams are already interested in replicating the current champion's dribbling.

As any change is itself a disadvantage, as teams need to change their software and already gathered knowledge is voided, I'd suggest dropping this pull request.

Instead, one could try to adopt the suggestion to increase the distance to 10 cm, as the autoref calls these infringements more strictly than a human referee usually did pre Japan.

tobiasheineken:
To further illustrate the point I made about work for teams and voiding previous experience:

I'd guess that we invest at least 15% of our development time a year to adopting our AI to the new rules. This does not include exploring new possibilities due to new rules, it's just adopting or removing already implemented behavior that's broken due changes in the rules.
And to me, this sounds like too much.

This rule change itself can be implemented easily in the Game controller or the auto refs.
Teams do not **have to** adjust their software for an AI that obeys the new rules.

However, simply obeying the rules isn't the whole story.
We aren't satisfied with a legal reply, we want to have a **good reply** for a given situation.
We decided that for defending a free kick - knowing that the robot may not touch the ball twice, and only triggering if no other opponent robot is nearby - it is best to accelerate tangential to the stop-circle that you may not enter during stop as soon as we're convinced the opponent will shoot immediately.
Due to the latency of the system, it might not be legal to enter the stop-circle, as we might start slightly earlier than the state-change is registered, and investing energy for staying on the circle was decided to be worse than pure tangential acceleration.

These thoughts are voided by this proposal:
We can no longer safely assume that the opponent doesn't use the free space to simply drive with the ball. Therefore, a new defense should be developed. We most likely have to stay closer to the attacking robot.

The same holds true for shooting the free kick ourselves. We implemented a detection to avoid double touch if possible. This code is no longer needed. Also, not changing anything doesn't explore any of the dozens new possibilities given to the attacker due to this rule change. Which sadly might be the correct approach to SSL right now: Do not invest time in exploring possibilities to improve your AI by looking into and changing or adopting small details. The always changing rules will void them earlier or later.

After all, I have to admit that this isn't by far the worst proposal for this year, so it might not be worth the comment I'm writing right now. But as the same general point - with different, new examples - can be applied to #3 (which is especially true as a rule change one or two years ago allowed a new approach to defend an indirect free kick by stating that an _attacking robot_ has to touch the ball before it can enter the goal, freeing one robot from defending the direct shot in some cases, as touching the goalie and entering the goal would no longer be a valid goal), and other suggestions like walls, etc.

g3force:
The point regarding the defense is a good catch. I suggest to postpone the change for now.

mickmack1213:
How about a really small change?
Increase the double touch distance to 10cm. It's probably to less space to effectively make an attack move other than kicking, so the overall offense and defense strategies won't change much or at all.

It just helps with the imprecision.

g3force:
My impression (but it's not backed by data) is: Double touch many occurs when the robot moves forward but the barrier is not triggering, either because the kicker is broken, or because the ball is on the corner of the kicking device. In both cases, the increased distance would probably not help.

Additionally, having a different distance for this rule and for the "ball is in play" for free kicks has new side effects, so it is not as simple as one would think :P

rolfvdhulst:
I agree with @tobiasheineken here
Also, does anyone have any proof for raising it to 10 cm? SSL-Vision's imprecision is at worst  roughly 4-5 cm and more realistically 1-2 cm for a ball with a radius of 2.15 cm (unless parts on the robots are being detected as balls). Particularly when the ball has been lying still and has been visible for >0.5 seconds, as is the case for most free kicks, it should be easy to determine whether or not the ball has moved 5 centimeters due to vision inaccuracy or due to an actual robot kicking it.

In particular for penalty kicks I would argue that 10 cm is too much. At that point almost all teams' software will have decided that e.g. a keeper should dive to intercept a ball. 10 cm will allow for abuse of the rule in that sense, forcing old teams to again adapt their code. I would not arbitrarily change the margin from 5 centimeters to 10 for vision imprecision when 5 has been shown to be sufficient in previous years.

PR #7, opened 2020-01-21
Add rule for changing keeper Id

leonardoSCosta:
From the Rule change proposal:

> The keeper ID can be automatically changed with the team protocol which was introduced in 2019, but teams could only change the keeper during STOP.
> This is an unnecessary limitation. With reduced number of stops, this becomes even more important.
> 
> We propose to allow changing the keeper id, as long as the ball is on the opponent's half. This applies to the team protocol as well as to the manual change via the game controller operator.

### DIFF
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -97,7 +97,9 @@ The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers
 ==== Choosing Keeper Id
 The <<Referee, referee>> asks both <<Robot Handler, robot handlers>> which robot they will use as the keeper and forwards this information to the <<Game Controller Operator, game controller operator>>.
 
-The keeper Id can also be changed during game play, this can be done automatically by using the <<Game Controller, game controller>> network interface or manually, in this case, the <<Game Controller Operator, game controller operator>> is responsible for checking if the ball is in a valid position and making the change.
+The keeper id can also be changed during game play. This can be done automatically by using the <<Game Controller, game controller>> network interface or by asking the <<Game Controller Operator, game controller operator>> to change it in the <<Game Controller, game controller>>. The <<Game Controller Operator, game controller operator>> must not change the keeper id until the ball is at a valid position.
+
+Teams can automatically change their keeper id when the ball is on the opponent's half by using the <<Game Controller, game controller>> network interface.
 
 NOTE: If a team does not want to use a keeper, it may select the id of a robot that is not on the field.
 
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -97,6 +97,8 @@ The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers
 ==== Choosing Keeper Id
 The <<Referee, referee>> asks both <<Robot Handler, robot handlers>> which robot they will use as the keeper and forwards this information to the <<Game Controller Operator, game controller operator>>.
 
+The keeper Id can also be changed during game play, this can be done automatically by using the <<Game Controller, game controller>> network interface or manually via the <<Game Controller Operator, game controller operator>>.
+
 NOTE: If a team does not want to use a keeper, it may select the id of a robot that is not on the field.
 
 === Game Stages
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -97,7 +97,7 @@ The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers
 ==== Choosing Keeper Id
 The <<Referee, referee>> asks both <<Robot Handler, robot handlers>> which robot they will use as the keeper and forwards this information to the <<Game Controller Operator, game controller operator>>.
 
-. If the ball is <<Ball In And Out Of Play, in play>>, the keeper id can only be changed if it is on the opponent's half. There is an automatic and a manual procedure, which are defined below, respectively:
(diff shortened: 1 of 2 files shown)


chapters/gamestructure.adoc
```diff
@@ -97,6 +97,8 @@ The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers
 ==== Choosing Keeper Id
 The <<Referee, referee>> asks both <<Robot Handler, robot handlers>> which robot they will use as the keeper and forwards this information to the <<Game Controller Operator, game controller operator>>.
 
+The keeper Id can also be changed during game play, this can be done automatically by using the <<Game Controller, game controller>> network interface or manually, in this case, the <<Game Controller Operator, game controller operator>> is responsible for checking if the ball is in a valid position and making the change.
```
g3force: `keeper id` (lower case)

g3force: ... play. This can be... (new sentence)

g3force: it is a bit unclear, what the last sentence is referring to. What about: `interface or by asking the GCO to change it in the GC. The GCO must not change the keeper id until the ball is at a valid position.`

chapters/playingenvironment.adoc
```diff
@@ -82,7 +82,7 @@ Besides the shared vision equipment, teams are not allowed to mount their own ca
 A game is controlled by the community-maintained ssl-game-controller (https://github.com/RoboCup-SSL/ssl-game-controller).
 It is operated by the <<Game Controller Operator, game controller operator>>. The software translates decisions of the <<Referee, referee>> and the <<Automatic Referee, automatic referee>> into Ethernet communication signals that are broadcast to the network. It maintains the state of the game, tracks all events and acts as a proxy between all participating parties in the game.
 
-The game-controller has a network interface for the playing teams. They can automatically change their keeper id when the ball is <<Ball In And Out Of Play, out of play>>, they can signal a robot substitution intent for the next opportunity and they can reply to requests of the <<Advantage Rule, advantage rule>>.
+The game-controller has a network interface for the playing teams. They can automatically change their keeper id when the ball is on the opponent's half, they can signal a robot substitution intent for the next opportunity and they can reply to requests of the <<Advantage Rule, advantage rule>>.
```
g3force: I'd remove `when the ball is on the opponent's half` and add this as a separate paragraph above in the choosing keeper id section, where I'd expect this rule more than here.

g3force (changes requested):

chapters/gamestructure.adoc
```diff
@@ -97,6 +97,10 @@ The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers
 ==== Choosing Keeper Id
 The <<Referee, referee>> asks both <<Robot Handler, robot handlers>> which robot they will use as the keeper and forwards this information to the <<Game Controller Operator, game controller operator>>.
 
+The keeper id can also be changed during game play. This can be done automatically by using the <<Game Controller, game controller>> network interface or by asking the <<Game Controller Operator, game controller operator>> to change it in the <<Game Controller, game controller>>. The <<Game Controller Operator, game controller operator>> must not change the keeper id until the ball is at a valid position.
+
+Teams can automatically change their keeper id when the ball is on the opponent's half by using the <<Game Controller, game controller>> network interface.
```
g3force: It is not clear, that this also refers to the manual change. What about:

> The keeper id may only be changed when ...

g3force: Or alternatively merge with the section before:

> The keeper id can also be changed during game play, as long as the ball is in the opponent's half. This can be done automatically by using the <<Game Controller, game controller>> network interface or by asking the <<Game Controller Operator, game controller operator>> to change it in the <<Game Controller, game controller>>.

Should be clear then, that the condition applies to both automatic and manual change.

Maybe add a note, that the team should only ask the GCO when the condition is met and that the GCO is responsible for complying to the rules. I think, that's the best procedure.

g3force (changes requested):

chapters/gamestructure.adoc
leonardoSCosta: I think I got it, see if it is clearer now

chapters/gamestructure.adoc
```diff
@@ -97,6 +97,12 @@ The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers
 ==== Choosing Keeper Id
 The <<Referee, referee>> asks both <<Robot Handler, robot handlers>> which robot they will use as the keeper and forwards this information to the <<Game Controller Operator, game controller operator>>.
 
+. If the ball is <<Ball In And Out Of Play, in play>>, the keeper id can only be changed if it is on the opponent's half. There is an automatic and a manual procedure, which are defined below, respectively:
```
```suggestion
. If the ball is <<Ball In And Out Of Play, in play>>, the keeper id can only be changed if the ball is in the opponent's half. There is an automatic and a manual procedure, which are defined below, respectively:
```

g3force (changes requested):

g3force (approved):

mickmack1213:
What's about two Bots in the defense area during the change? 
Are both allowed to touch the ball? For how many seconds?

I have to admit it's quite unlikely to be important but maybe think about it.

g3force:
The keepers may only be switched when the ball is in the opponent half, to avoid the situation you mentioned. No grace period needed. The keeper id is changed in the GC and afterwards the old keeper may not touch the ball in the defense area.

PR #9, opened 2020-01-21
Robot interchange during game play (Both Div)

leonardoSCosta:
From the Rule proposal:

> We aim for less stoppages and especially with walls, there will probably be much less chances to interchange robots. We already encourage teams to automatically interchange robots by replacing them at the midline, but the game still had to be stopped and even halted in 2019. This wasted overall game time.
> 
> We propose that robots can always be taken in and out from the midline. No need to stop the game. There should not be any action at the midline during the replacement. That could be enforced by requiring that the ball is not near the interchange position.

> In 2019, a lot of time was lost in HALT, while teams interchanged their robots.
> We want to encourage teams to reduce the number of manual interchanges to a minimum and make use of the automatic interchange to save overall game time.
> 
> We propose that manual robot interchange (if the robot is not taken in/out from the midline) always requires a timeout to be taken.
> If a team is out of timeouts and still needs to take one robot out, the team is allowed to take the robot out without a timeout, but can not use any remaining timeout time. It can only remove the robot from the field as fast as possible.
>

### DIFF
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -72,7 +72,7 @@ Before the start of the match, every team has to designate one robot handler. Th
 .Duties
 * The robot handler helps <<Match Preparation, preparing the match>>.
 * The robot handler asks the referee for <<Timeouts, timeouts>> if necessary.
-* The robot handler can substitute a robot during game play, if it is at the halfway line (Division A only), <<Robot Substitution, substitutes the robot>>.
+* The robot handler can substitute a robot during game play, <<Robot Substitution, substitutes the robot>>.
 * The robot handler asks the referee for the permission to substitute a robot in the next stoppage and, if the referee agrees, <<Robot Substitution, substitutes the robot>>.
 * The robot handler voices concerns of the team (for example network issues or vision problems).
 
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -72,7 +72,7 @@ Before the start of the match, every team has to designate one robot handler. Th
 .Duties
 * The robot handler helps <<Match Preparation, preparing the match>>.
 * The robot handler asks the referee for <<Timeouts, timeouts>> if necessary.
-* The robot handler can substitute a robot during game play, <<Robot Substitution, substitutes the robot>>.
+* The robot handler can <<Robot Substitution, substitute a robot during game play>>.
 * The robot handler asks the referee for the permission to substitute a robot in the next stoppage and, if the referee agrees, <<Robot Substitution, substitutes the robot>>.
 * The robot handler voices concerns of the team (for example network issues or vision problems).
 
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -72,7 +72,7 @@ Before the start of the match, every team has to designate one robot handler. Th
 .Duties
 * The robot handler helps <<Match Preparation, preparing the match>>.
 * The robot handler asks the referee for <<Timeouts, timeouts>> if necessary.
-* The robot handler can substitue a robot during game play, if it is at the midline (Division A only), <<Robot Substitution, substitutes the robot>>.
+* The robot handler can substitute a robot during game play, if it is at the halfway line (Division A only), <<Robot Substitution, substitutes the robot>>.
 * The robot handler asks the referee for the permission to substitute a robot in the next stoppage and, if the referee agrees, <<Robot Substitution, substitutes the robot>>.
 * The robot handler voices concerns of the team (for example network issues or vision problems).
 
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -72,6 +72,7 @@ Before the start of the match, every team has to designate one robot handler. Th
 .Duties
 * The robot handler helps <<Match Preparation, preparing the match>>.
 * The robot handler asks the referee for <<Timeouts, timeouts>> if necessary.
+* The robot handler can substitue a robot during game play, if it is at the midline (Division A only), <<Robot Substitution, substitutes the robot>>.
 * The robot handler asks the referee for the permission to substitute a robot in the next stoppage and, if the referee agrees, <<Robot Substitution, substitutes the robot>>.
 * The robot handler voices concerns of the team (for example network issues or vision problems).
 
--- a/chapters/substitution.adoc
+++ b/chapters/substitution.adoc
@@ -2,18 +2,22 @@
 .Definition
 Robots are substituted by the <<Robot Handler, robot handler>> of the respective team. No other team member is allowed to take robots out or put robots in.
 
-. The <<Robot Handler, robot handler>> must preferably not wear colors that may interfere with the vision system.
-.. If non-stop robot interchange is allowed (Division A), the <<Robot Handler, robot handler>> should preffer using long sleeves.
+. Robot substitution during game play - Division A only:
+.. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
+... Put robots in during game play. If the ball is at least 2 meters away from the mid line, the robot can be placed in it without notifying the <<Referee, referee>>.
+... Take robots out during game play (Division A only). If a robot is at the mid line and the ball is at least 2 meters away from the mid line, the robot can be removed without notifying the <<Referee, referee>>.
 
-. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
-.. Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
-.. Put robots in during game play (Division A only). If the ball is at least 2 meters away from the midline, the robot can be placed in it without notifying the <<Referee, referee>>.
-.. Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
-.. Take robots out during game play (Division A only). If a robot is at the midline and the ball is at least 2 meters away from the midline, the robot can be removed without notifying the <<Referee, referee>>.
+. Robot substitution on next stoppage - Both Divisions:
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
+... Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
+... Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
+... The <<Robot Handler, robot handler>> informs the <<Referee, referee>> when done.
+... When both teams finished the robot substitution, the <<Referee, referee>> informs the <<Game Controller Operator, game controller operator>>.
+... The <<Game Controller Operator, game controller operator>> performs a <<Stop, stop>> followed by continuing the game.
+
 . If one of the robots taken out is the keeper, the <<Robot Handler, robot handler>> tells the <<Game Controller Operator, game controller operator>> the id of the robot that takes over the keeper role.
-. The <<Robot Handler, robot handler>> informs the <<Referee, referee>> when done.
-. When both teams finished the robot substitution, the <<Referee, referee>> informs the <<Game Controller Operator, game controller operator>>.
-. The <<Game Controller Operator, game controller operator>> performs a <<Stop, stop>> followed by continuing the game.
+
 
 .Usage
 Robots can be substituted for any reason. There is no limit on the number of substitutions.
@@ -23,6 +27,7 @@ A robot substitution intent can be made by:
 . A <<Robot Handler, robot handler>> by informing the <<Game Controller Operator, game controller operator>> who in turn enters the intent into the <<Game Controller, game controller>>.
 . A team software by sending a request to the <<Game Controller, game controller>>.
 . The <<Game Controller, game controller>> itself if a team exceeds the maximum number of robots (for example after a team receives a <<Yellow Card, yellow>> or <<Red Card, red card>>).
+. *(Division A only)* If a robot needs to be taken out and it is not in the mid line, a timeout needs to be taken in order for the team to remove the robot from the field. If a team is out of timeouts, it will still be able to take the robot out, however, it will not be able to use any remaining timeout time, the robot must be removed from the field as fast as possible.
 
 If the game was halted due to a substitution intent by a team, at least one substitution (taking a robot out or putting one in) must be performed by this team. A substitution intent can be revoked unless the game was not already halted for substitution.
 
--- a/chapters/substitution.adoc
+++ b/chapters/substitution.adoc
@@ -4,12 +4,15 @@ Robots are substituted by the <<Robot Handler, robot handler>> of the respective
 
 . Robot substitution during game play:
 .. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
-.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
-... Put robots in. If the ball is at least 2 meters away from the halfway line, the robot can be placed on the intersection of the halfway line with any touch line without notifying the <<Referee, referee>>.
-... Take robots out. If a robot is at the intersection of the halfway line with any touch line and the ball is at least 2 meters away from the halfway line, the robot can be removed without notifying the <<Referee, referee>>.
+.. A robot can only be substituted while obeying the following specifications:
+... Is at least on top of any touch line, if possible the robot can also be placed beyond the field boundaries
+... At a distance from the halfway line that must not exceed 30 centimeters
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs the following operations.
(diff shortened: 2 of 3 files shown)


leonardoSCosta:
To keep the consistency with the do not stop for yellow cards rule #10, we've decided that the automatic interchange will also apply to Div B. Note that this isn't something that is mandatory to be used, but, it is something that can help those who want to use it

chapters/substitution.adoc
```diff
@@ -2,13 +2,22 @@
...
-.. Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
-.. Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
+. Robot substitution during game play:
+.. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
+... Put robots in. If the ball is at least 2 meters away from the halfway line, the robot can be placed on the intersection of the halfway line with any touch line without notifying the <<Referee, referee>>.
+... Take robots out. If a robot is at the intersection of the halfway line with any touch line and the ball is at least 2 meters away from the halfway line, the robot can be removed without notifying the <<Referee, referee>>.
```
MathewMacDougall: Is there a threshold for how close the robot must me to the intersection of the halfway line and touch line like before? Or in the case of "live" substitutions are robots now expected to be in the nearly exact position?

chapters/substitution.adoc
```diff
@@ -2,13 +2,22 @@
...
+
+. Robot substitution on next stoppage:
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
+... Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
```
MathewMacDougall: This is a bit of a nitpick, but if the robot handler follows these steps in order they may temporarily exceed the maximum number of robots of the team on the field because one must be added before another can be removed. Perhaps we should clarify the maximum allowed number of robots of the team on the field must not be exceeded _when the substitutions are finished_? Unless this is intended.

MathewMacDougall (changes requested):

chapters/substitution.adoc
leonardoSCosta: I don't think this is intended, I believe that what it means is that the robot handler can perform any of those actions, not necessarily in order. Doesn't make much sense only being able to take a robot out if you put one in. It is the same idea as previous years, there can never be more robots than the allowed for each team, and, there's nothing that prohibits having less robots. I'll fix this.

chapters/substitution.adoc
leonardoSCosta: I'll specify it better. I think that the robot doesn't need to be exactly on the intersection, what is most important is that the robot is at least on top of the touch line, to avoid any interference with the vision system, and it could be at 30cm ish from the halfway line.

MathewMacDougall (approved):
:+1:

chapters/gamestructure.adoc
```diff
@@ -72,6 +72,7 @@ Before the start of the match, every team has to designate one robot handler. Th
 .Duties
 * The robot handler helps <<Match Preparation, preparing the match>>.
 * The robot handler asks the referee for <<Timeouts, timeouts>> if necessary.
+* The robot handler can substitute a robot during game play, <<Robot Substitution, substitutes the robot>>.
```
g3force: that's not a valid sentence, I guess? :D

chapters/substitution.adoc
```diff
@@ -18,6 +30,7 @@ A robot substitution intent can be made by:
 . A <<Robot Handler, robot handler>> by informing the <<Game Controller Operator, game controller operator>> who in turn enters the intent into the <<Game Controller, game controller>>.
 . A team software by sending a request to the <<Game Controller, game controller>>.
 . The <<Game Controller, game controller>> itself if a team exceeds the maximum number of robots (for example after a team receives a <<Yellow Card, yellow>> or <<Red Card, red card>>).
+. *(Division A only)* If a robot needs to be taken out and it is not in the mid line, a timeout needs to be taken in order for the team to remove the robot from the field. If a team is out of timeouts, it will still be able to take the robot out, however, it will not be able to use any remaining timeout time, the robot must be removed from the field as fast as possible.
```
g3force: Can we put this into a separate PR? This has a lot more consequences for teams than the rest and would be a candidate for postponement to 2021.

chapters/substitution.adoc
```diff
@@ -2,13 +2,25 @@
...
-.. Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
-.. Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
+. Robot substitution during game play:
+.. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+.. A robot can only be substituted while obeying the following specifications:
+... Is at least on top of any touch line, if possible the robot can also be placed beyond the field boundaries
+... At a distance from the halfway line that must not exceed 30 centimeters
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs the following operations.
+... Put robots in. If the ball is at least 2 meters away from the halfway line, the robot can be substituted without notifying the <<Referee, referee>>.
```
g3force: can you try to remove the 2 meters away definition duplication?

chapters/substitution.adoc
```diff
@@ -2,13 +2,25 @@
...
+... Take robots out. If the robot is placed at the previous specified position and the ball is at least 2 meters away from the halfway line, the robot can be removed without notifying the <<Referee, referee>>.
+
+. Robot substitution on next stoppage:
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs the following operations.
+... Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
+... Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
```
g3force: With the new rules, "the next opportunity" does not make sense anymore, because you can now add robots whenever you want.

The rule was added to motivate the automatic interchange. With new changes, there is an even higher motivation for this, because you can substitute robots immediately, so I think it is fine to drop this line.

chapters/substitution.adoc
```diff
@@ -2,13 +2,25 @@
 .Definition
 Robots are substituted by the <<Robot Handler, robot handler>> of the respective team. No other team member is allowed to take robots out or put robots in.
 
-. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
-.. Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
-.. Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
+. Robot substitution during game play:
+.. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+.. A robot can only be substituted while obeying the following specifications:
+... Is at least on top of any touch line, if possible the robot can also be placed beyond the field boundaries
```
g3force: Suggestion: Is at least partially inside the field margin

but double check if the field margin is defined as such. Maybe add it into the field drawings?

g3force (changes requested):
The definition of the interchange has some duplications, both in the new part and with the existing rules.

I find it confusing that for the interchange during stop, there is a 1m radius, while during play there is another rule.

I would remove the old definition with the 1m radius and replace it by the new one, except that during stop, robots can be removed from everywhere on the field.

We can further optimize the procedure:
During stop, robots can be removed.
Robots can always be added (given the conditions for substitution during game).
I think, that way the whole section can hugely be simplified.

chapters/substitution.adoc
leonardoSCosta: Makes sense, but what if a team still wants to manually replace a robot? For example, a robot that for some reason got a low battery and cannot move, it won't cause any damage to itself or other robots, thus, a emergency stop is not required, but it will still need to be manually removed from the field, this kind of situation can wait until a game stoppage

leonardoSCosta: The 1 meter rule doesn't make much sense, since now it is possible to put robots in at any time, so I'll just remove it

chapters/substitution.adoc
leonardoSCosta: Done. PR #16

chapters/substitution.adoc
g3force: Without #16, teams can still substitute robots during stop as usual. No need for an emergency stop.

chapters/substitution.adoc
```diff
@@ -2,21 +2,36 @@
...
 Robots are substituted by the <<Robot Handler, robot handler>> of the respective team. No other team member is allowed to take robots out or put robots in.
 
-. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
-.. Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
-.. Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
+. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+
+. Robots can always be substituted during game play if the following specifications are respected:
+
+.. Is at least partially inside the <<Field Surface, field margin>>
```
g3force: > The robot is as least...

chapters/substitution.adoc
```diff
@@ -2,21 +2,36 @@
...
+.. At a distance from the halfway line that must not exceed 30 centimeters
```
g3force: Prefer SI units here, so 0.3m.

What was the rational again for 0.3m? We should consider that there might be both robot handlers on the same side, as we gave no limitations as to where they should be. Last year, we asked them to be both on the opposite side from the GCO.

To me, 1m would sound just as fine.

chapters/substitution.adoc
```diff
@@ -2,21 +2,36 @@
...
+.. The ball must be at least 2 meters away from the halfway line
+
+. If the previous specifications are respected, the <<Robot Handler, robot handler>> is allowed to perform the following operations.
+... Put robots in, without notifying the <<Referee, referee>>.
+... Take robots out, without notifying the <<Referee, referee>>.
+
+. Given that robots can be put in at any time, while respecting the previous specifications, on a game stoppage robots can only be removed.
```
g3force: That sounds a bit complicated. What about:

> Additionally, robots can be taken out from any position during game stoppage

chapters/substitution.adoc
```diff
@@ -2,21 +2,36 @@
...
+. The maximum allowed number of robots of the team on the field must not be exceeded.
+
 
 .Usage
 Robots can be substituted for any reason. There is no limit on the number of substitutions.
 
 A robot substitution intent can be made by:
 
 . A <<Robot Handler, robot handler>> by informing the <<Game Controller Operator, game controller operator>> who in turn enters the intent into the <<Game Controller, game controller>>.
-. A team software by sending a request to the <<Game Controller, game controller>>.
```
g3force: Why is this removed?

chapters/substitution.adoc
```diff
@@ -2,21 +2,36 @@
...
-.. Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
+. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+
+. Robots can always be substituted during game play if the following specifications are respected:
+
+.. Is at least partially inside the <<Field Surface, field margin>>
+.. At a distance from the halfway line that must not exceed 30 centimeters
+.. The ball must be at least 2 meters away from the halfway line
+
+. If the previous specifications are respected, the <<Robot Handler, robot handler>> is allowed to perform the following operations.
```
g3force: Originally, the rules were defined with the order and the maximum number of robots in mind, that's why there were those two options (put in, take out). The idea was: The robot handler can put robots in until the maximum number is reached. Then the robot handler can take robots out again. That way, a team had to play with one robot less until the next stoppage.

Now in the new context, this formulation does not make much sense to me anymore. I would just simplify it to one sentence.

chapters/substitution.adoc
```diff
@@ -2,21 +2,36 @@
...
+... Put robots in, without notifying the <<Referee, referee>>.
+... Take robots out, without notifying the <<Referee, referee>>.
+
+. Given that robots can be put in at any time, while respecting the previous specifications, on a game stoppage robots can only be removed.
+.. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs the following operation in the given order.
```
g3force: Merge this with the point before?

chapters/substitution.adoc
```diff
@@ -2,21 +2,36 @@
...
+... The <<Robot Handler, robot handler>> informs the <<Referee, referee>> when done.
+... When both teams finished the robot substitution, the <<Referee, referee>> informs the <<Game Controller Operator, game controller operator>>.
+... The <<Game Controller Operator, game controller operator>> performs a <<Stop, stop>> followed by continuing the game.
+
 . If one of the robots taken out is the keeper, the <<Robot Handler, robot handler>> tells the <<Game Controller Operator, game controller operator>> the id of the robot that takes over the keeper role.
-. The <<Robot Handler, robot handler>> informs the <<Referee, referee>> when done.
-. When both teams finished the robot substitution, the <<Referee, referee>> informs the <<Game Controller Operator, game controller operator>>.
-. The <<Game Controller Operator, game controller operator>> performs a <<Stop, stop>> followed by continuing the game.
+
+. The maximum allowed number of robots of the team on the field must not be exceeded.
```
g3force: > ... at any time

g3force (changes requested):

chapters/substitution.adoc
leonardoSCosta: With 0.3 m there would be a total of 0.6 m (both sides of the halfway line), I believe that's enough for more than one robot being substituted at the same time, but I'm okay with 1 m too

chapters/substitution.adoc
leonardoSCosta: Yeah, I agree, thanks :sweat_smile:

chapters/substitution.adoc
leonardoSCosta: With the new automatic procedure I didn't see why this would be necessary. Should it be kept?

g3force: if a robot is broken and must be taken off the field, you still want to request substitution.

chapters/substitution.adoc
```diff
@@ -2,13 +2,25 @@
...
+. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+
+. Robots can always be substituted during game play if the following specifications are respected:
+
+.. The robot is at least partially inside the <<Field Surface, field margin>>
+.. The robot is at a distance from the halfway line that must not exceed 1 meter
+.. The ball must be at least 2 meters away from the halfway line
+
+. If the previous specifications are respected, the <<Robot Handler, robot handler>> is allowed to put robots in and take robots out, without notifying the <<Referee, referee>>.
+.. Additionally, robots can be taken out from any position during game stoppage by the <<Robot Handler, robot handler>> using the procedure below:
```
g3force: I think there is no need for the extra indents here, as there is only one point on the second level (this line).

Could you btw. add a reference for game stoppage here? :)

chapters/substitution.adoc
```diff
@@ -2,13 +2,25 @@
...
 Robots are substituted by the <<Robot Handler, robot handler>> of the respective team. No other team member is allowed to take robots out or put robots in.
 
-. A <<Robot Handler, robot handler>> who's team is marked for robot substitution performs at least one of the following operations in the given order.
-.. Put robots in. The maximum allowed number of robots of the team on the field must not be exceeded.
-.. Take robots out. If a robot that is taken out is closer than 1 meter to the intersection of the halfway line with one of the touch lines, another robot can be put in right away. Otherwise the robot handler has to wait until the next substitution opportunity.
+. The <<Robot Handler, robot handler>> should prefer to use long sleeves and colors that won't interfere with the vision system.
+
+. Robots can always be substituted during game play if the following specifications are respected:
+
+.. The robot is at least partially inside the <<Field Surface, field margin>>
```
g3force: General note: It seems that most of the bullet point lists end with a dot. It would be nice to keep consistency.

chapters/substitution.adoc
```diff
@@ -2,13 +2,25 @@
...
+.. The robot is at a distance from the halfway line that must not exceed 1 meter
+.. The ball must be at least 2 meters away from the halfway line
+
+. If the previous specifications are respected, the <<Robot Handler, robot handler>> is allowed to put robots in and take robots out, without notifying the <<Referee, referee>>.
```
g3force: Sry for nitpicking, but this is sentence seems partially redundant. You already said before, that the specifications have to be respected and now you repeat it. I would just add the "without notifying to the point above:

> Robots can always be substituted during game play without notifying the <<Referee, referee>> if the following specifications are respected:

g3force (changes requested):

chapters/substitution.adoc
leonardoSCosta: No problem, if you find anything that could be improved just tell me

g3force (approved):

PR #10, opened 2020-01-23
Do not stop for cards (both Div)

leonardoSCosta:
From the rule changes proposal:

> With the introduction of the autoRef, a lot more fouls were detected then before. This also results in more yellow cards, which in turn stop the game. 
> 
> We propose to not stop the game on yellow cards immediately, but allow teams within 10s to move a robot to the interchange position, where it can be taken out immediately.
> This is especially useful for bot crashes, which increase the foul counter, which in turn may add yellow cards.
> If a team did not get a robot out within time, the game is stopped and the ball is placed in front of the teams defense area and the other team gets a freekick.
> 
> A team can't score a goal while having more than the allowed number of robots on the field.


chapters/refereecommands.adoc
```diff
@@ -148,11 +148,17 @@ Penalty Kicks are used to punish teams that received multiple <<Yellow Card, yel
...
 ==== Yellow Card
 .Definition
-A yellow card can only be given during <<Halt, halt>>.
 
 If the yellow card is shown as a result of <<Unsporting Behavior, unsporting behavior>>, the referee may decide to immediately <<Halt, halt>> the match. In this case, the match continues with a direct free kick for the other team.
 
-Upon receipt of a yellow card, the number of robots allowed on the field for the penalized team decreases by one. If, after this decrease, the team has more robots than permitted on the field, a robot must be <<Robot Substitution, taken out>> before <<Resuming The Game, play resumes>>. The penalized team can choose the robot to remove.
+Upon receipt of a yellow card, the number of robots allowed on the field for the penalized team decreases by one. If, after this decrease, the team has more robots than permitted on the field, a robot must be <<Robot Substitution, taken out>>.
+
+After the yellow card is issued the game will not be stopped, the team will have 10 seconds to <<Robot Substitution, automatically remove the robot>>. If a robot is not taken out within time, the game is stopped for <<Robot Substitution, manual substitution>>.
```
g3force: Let's add a note here, that the game will still be stopped for fouls, if those fouls cause a stop.

Especially the rule for dropping parts will still cause a stop, so teams are able to replace the robot.

chapters/refereecommands.adoc
```diff
@@ -148,11 +148,19 @@ Penalty Kicks are used to punish teams that received multiple <<Yellow Card, yel
...
+
+NOTE: There are some fouls that causes game stoppage, e.g. dropping parts. Therefore, if any of these fouls happen within 10 seconds of the yellow card, teams are allowed to replace the robot.
```
g3force: that's not, what I meant :D

A yellow card can be given for different reasons, some of which do not cause the game to be stopped, e.g. multiple fouls, where the last foul was a robot crash, which does not stop the game. But if the last foul that caused the yellow card (possibly indirectly through multiple fouls) is for example robot tipping over, than the game is obviously be stopped (actually even before the yellow card is issues).

chapters/refereecommands.adoc
```diff
@@ -148,11 +148,19 @@ Penalty Kicks are used to punish teams that received multiple <<Yellow Card, yel
...
 ==== Yellow Card
 .Definition
-A yellow card can only be given during <<Halt, halt>>.
 
 If the yellow card is shown as a result of <<Unsporting Behavior, unsporting behavior>>, the referee may decide to immediately <<Halt, halt>> the match. In this case, the match continues with a direct free kick for the other team.
 
-Upon receipt of a yellow card, the number of robots allowed on the field for the penalized team decreases by one. If, after this decrease, the team has more robots than permitted on the field, a robot must be <<Robot Substitution, taken out>> before <<Resuming The Game, play resumes>>. The penalized team can choose the robot to remove.
+Upon receipt of a yellow card, the number of robots allowed on the field for the penalized team decreases by one. If, after this decrease, the team has more robots than permitted on the field, a robot must be <<Robot Substitution, taken out>>.
+
+After the yellow card is issued the game will not be stopped, the team will have 10 seconds to <<Robot Substitution, automatically remove the robot>>. If a robot is not taken out within time, the game is stopped for <<Robot Substitution, manual substitution>>.
```
g3force: I would rather formulate it like 

> A yellow card does not lead to a stop automatically. If the the ball is <<in play>>, the team will have....

g3force (changes requested):

chapters/refereecommands.adoc
leonardoSCosta: Got it

g3force (approved):

g3force:
I think this rule should also be applied to red cards (which can be caused by multiple yellow cards, see #15). The red card already extends from yellow cards, so the rule definitions are already correct.

PR #11, opened 2020-01-24
Emergency stop (Div A)

leonardoSCosta:
From the rule change proposal:

> With less stoppages, there are fewer chances to get a robot from the field manually. Even with a timeout, the team would need to wait until the next stoppage.
> If something goes really wrong with a teams robot (like damaging itself more and more), the team might want to stop the game to reduce the damage (and any cost due to destroyed parts).
> 
> Proposal: A team can ask to stop the game immediately after a grace period of 10s, regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field as fast as possible, but can not use any remaining timeout time

### DIFF
--- a/chapters/emergencystop.adoc
+++ b/chapters/emergencystop.adoc
@@ -3,29 +3,24 @@
 This rule only applies to Division A.
 
 .Definition
+A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation.
+It will receive a yellow card for this and must take a timeout immediately.
+If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
 
-A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
-
-NOTE: This rule is supposed to only be used in extreme situations, e.g. a software crash or the robots starts damaging themselves.
+NOTE: This rule is supposed to be used in extreme situations only, e.g. a software crash or when robots are damaging themselves significantly.
 
 When the game is stopped due to this rule, there are three possibilities that may have happened:
 
-. 10 seconds have passed and the game is stopped.
-. The game is stopped earlier than 10 seconds due to the ball leaving the field or a foul occurred.
-. The human referee stopped the game manually earlier than 10 seconds.
+. The grace period has passed and the game is stopped.
+. The human referee stopped the game earlier.
+. The game is stopped earlier due to the ball leaving the field or because of a foul.
 
 For these possibilities there are two methods to proceed the game:
 
-. For *1* and *3*, the game is continued with a free kick for the opposing team.
-. For *2*, the game is continued like after a regular timeout.
+. For *1* and *2*, the game is continued with a <<Free Kick, free kick>> for the opposing team.
+. For *3*, the game is continued like after a regular timeout.
 
 .Usage
+An emergency stop intent can be made using <<Communication Flags, communication flags>>.
 
-An emergency stop intent can be made by:
-
-// To inform the referee a communication flag shall be used
-// TODO: link "informing" with the communication flag section
-
-. A <<Robot Handler, robot handler>> by informing the <<Referee, referee>> who shall stop the game within 10 seconds.
-
-NOTE: The Official Team Remote Control is just an option, if it is not available, the referee is allowed to stop the game earlier than 10 seconds, if there isn't an important play in action, e.g. no team is at the risk of scoring a goal or the ball is close to the halfway line.
+NOTE: The referee may to stop the game earlier if there is no promising play in action.
--- a/chapters/emergencystop.adoc
+++ b/chapters/emergencystop.adoc
@@ -1,0 +1,13 @@
+== Emergency stop
+
+This rule only applies to Division A.
+
+.Definition
+
+A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
+
+.Usage
+
+A emergency stop intent can be made by:
+
+. A <<Robot Handler, robot handler>> by informing the <<Game Controller Operator, game controller operator>> who in turn enters the intent into the <<Game Controller, game controller>>.
--- a/chapters/emergencystop.adoc
+++ b/chapters/emergencystop.adoc
@@ -6,6 +6,17 @@ This rule only applies to Division A.
 
 A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
 
+There are three possible situations when this rule is used:
+
+. 10 seconds have passed and the game is stopped;
+. The game is stopped earlier than 10 seconds due to the ball leaving the field or a foul occurred;
+. The human referee stopped the game manually earlier than 10 seconds.
+
+And there are two methods of restarting the game:
+
+. For situation 1 and 3, the game is continued with a free kick for the opposing team;
+. For situation 2, the game is continued like a regular timeout.
+
 .Usage
 
 An emergency stop intent can be made by:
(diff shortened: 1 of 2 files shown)


chapters/emergencystop.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+This rule only applies to Division A.
+
+.Definition
+
+A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
+
+.Usage
+
+A emergency stop intent can be made by:
```
g3force: Should be `An emergency stop`.

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+// To inform the referee a communication flag shall be used
+// TODO: link "informing" with the communication flag section
+
+. A <<Robot Handler, robot handler>> by informing the <<Referee, referee>> who shall stop the game within 10 seconds or at the next stoppage.
+
+NOTE: Until we implement an automatic procedure, e.g. Raspberry PI, the referee is allowed to stop the game earlier than 10 seconds, if there isn't a important play in action, e.g. no team is at the risk of scoring a goal or the ball is close to the halfway line.
```
g3force: `an important play`

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,18 @@
+== Emergency stop
+
+This rule only applies to Division A.
+
+.Definition
+
+A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
```
tobiasheineken: If I read this correctly, this Emergency stop starts a timeout for the team.

I'm pretty sure - although I cannot find it right now - that a time (in stop state) always ended with the previous command, either indirect or direct for yellow or blue.

Maybe a clarification here (and in #10) how to continue the game after the forced timeout without a resolution like in the stop-state version. How did we handle Yellow cards before #10, when they stopped a running game?

g3force: Good point. A regular timeout can only be taken during stop, so the game could be continued normally afterwards. Now, there are three possible situations:

1. The 10 seconds run off and the game is stopped.
2. The game is stopped before the 10 seconds run off because the ball left the field, a foul occurred, etc.
3. The human ref stopped the game manually before the 10 seconds run off

Suggested Solution 1:
For 2, we continue just like with a regular timeout.
For 1 and 3, we start with a a) force start or with a b) free kick for the opposing team.

Suggested Solution 2:
We continue with either a) force start or b) free kick for all three situations.

I tend towards solution 1b). Otherwise, the opposing team could do a foul within the 10s that is not punished with a freekick.

Regarding the yellow cards: I think, a yellow card should have been given only during stop. At least, that was the assumption, but I remember that there were inconsistencies in the GC in this regard, because collisions could cause fouls which in turn could result in multiple fouls violation which lead to a yellow card. Btw: The rule change proposal already specifies what to do in the case, the robots are not removed in time.

tobiasheineken: #10 states that "in 2020 no penalty will be given to the team [...] ". So there's no description how to continue the game after the game-state has been left forcefully due to 10 seconds running off the clock.

tobiasheineken: I'd go for 1a, but I think anything in Solution 1 sounds fair. I just want it written down in the rules.

leonardoSCosta: Just added these procedures to restart the game

> Suggested Solution 1:
>For 2, we continue just like with a regular timeout.
>For 1 and 3, we start with a ~~a) force start or with a~~ b) free kick for the opposing team.

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+There are three possible situations when this rule is used:
```
tobiasheineken: This is misleading. The rule can be used "whenever". These three options are results that could happen if the rule is used and how to proceed with the game afterwards

tobiasheineken (changes requested):

chapters/emergencystop.adoc
leonardoSCosta: I see, I'll re phrase it

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,29 @@
...
+When the game is stopped due to this rule, there are three possibilities that may have happened:
+
+. 10 seconds have passed and the game is stopped;
+. The game is stopped earlier than 10 seconds due to the ball leaving the field or a foul occurred;
+. The human referee stopped the game manually earlier than 10 seconds.
+
+For these possibilities there are two methods to proceed the game:
+
+. For *1* and *3*, the game is continued with a free kick for the opposing team;
+. For *2*, the game is continued like a regular timeout.
```
g3force: like after a regular timeout

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+.Usage
+
+An emergency stop intent can be made by:
+
+// To inform the referee a communication flag shall be used
+// TODO: link "informing" with the communication flag section
+
+. A <<Robot Handler, robot handler>> by informing the <<Referee, referee>> who shall stop the game within 10 seconds or at the next stoppage.
```
g3force: stop the game at the next stoppage? :D If the emergency stop is requested, the game should not be stopped later than the 10s

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+NOTE: Until we implement an automatic procedure, e.g. Raspberry PI, the referee is allowed to stop the game earlier than 10 seconds, if there isn't an important play in action, e.g. no team is at the risk of scoring a goal or the ball is close to the halfway line.
```
g3force: I would not speak as "we" here, but formulate it in a passive way.

Also, the Raspberry PI should better be called sth like "Official Team Remote Control".

I suggest to formulate it in a way, that the remote control is always only an additional allowed option, because we never know if it will work and it might not be available on regional events.

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+This rule only applies to Division A.
+
+.Definition
+
+A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
+
+When the game is stopped due to this rule, there are three possibilities that may have happened:
+
+. 10 seconds have passed and the game is stopped;
```
g3force: are those `;` at the end there by intention?

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,29 @@
+== Emergency stop
+
+This rule only applies to Division A.
+
+.Definition
+
+A team can ask to stop the game immediately after a grace period of 10 seconds or at the next stoppage, whichever happens first regardless of the current situation. It will receive a yellow card for this and must take a timeout immediately. If the team is out of timeouts, it is still allowed to remove robots from the field, but can not use any remaining timeout time.
```
g3force: An info box explaining the rational of this rule might be a good idea. Stating that this is intended for fatal situations, e.g. where robots damage themselves.

g3force (changes requested):
Added some comments.

Remember that there is also still a TODO.

chapters/emergencystop.adoc
leonardoSCosta: Oops :smile:

chapters/emergencystop.adoc
leonardoSCosta: Okay, then we probably need to define this in the playing environment section, right?

g3force: yep

leonardoSCosta: Would you mind doing it? Since you have a better idea of what features it will have and how it will work

g3force: ok

g3force: I've edited this PR and the communication flags PR.

g3force (approved):

tobiasheineken (approved):

chapters/emergencystop.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+For these possibilities there are two methods to proceed the game:
+
+. For *1* and *2*, the game is continued with a <<Free Kick, free kick>> for the opposing team.
+. For *3*, the game is continued like after a regular timeout.
+
+.Usage
+An emergency stop intent can be made using <<Communication Flags, communication flags>>.
+
+NOTE: The referee may to stop the game earlier if there is no promising play in action.
```
g3force:
```suggestion
NOTE: The referee may stop the game earlier if there is no promising play in action.
```

PR #14, opened 2020-02-01
Remove Penalty Kicks (both Div)

g3force:
From the rule proposal:

> In 2019, there are two ways to receive a penalty kick:
> * Every third card led to a penalty kick
> * If a defender other than the keeper is fully inside the defense area and touches the ball
> 
> Penalty kicks have several disadvantages:
> * They are quite unbalanced (and thus not very interesting)
> * They lack innovation
> * They need extra implementation
> * They require additional time during a match
> * They interrupt the game
> 
> We propose to get rid of penalty kicks completely:
> * The Multiple Defenders foul will result in a red card. The game will be stopped immediately.
> * The third active yellow card will be turned into a red card instead. A yellow card is active as long as the counter has not run out, so in order to get a red card, the team must collect three yellow cards within 2 minutes.

During the last team meeting, it was suggested to replace the current penalty kick by the one-on-one procedure that is currently only applied in the shoot-out phase.
This is not as hard of a punishment as the current penalty kick, but it was found a better punishment than the red card.

This PR replaces the penalty kick procedure by the one-on-one procedure. There are no changes in what causes a penalty kick in this PR.


chapters/refereecommands.adoc
```diff
@@ -128,18 +128,33 @@ It can also be used to resume the game when the game had to be stopped and no te
...
+* the ball touches the inner surface of a goal wall or the ground of the goal of the defending team in less than 10 seconds, starting from when the <<Normal Start, normal start>> command is issued
+* the defending team violates any rule
+
+A goal is not awarded if:
+
+* the ball crosses any <<Field Lines, field lines>> outside the goal
+* the defending keeper touches the ball such that the ball speed vector changes direction by at least 90 degrees in 2D space
+* the attacking team violates any rule
+
+NOTE: The restrictions defined for <<Scoring Goals, scoring goals>> do not apply here. This explicitly includes the ball height limit of 0.15 meters. Other rules like the <<Excessive Dribbling, excessive dribbling>> limitation for example do.
```
MathewMacDougall: Just to be extra clean, the restriction of the ball height limit of 0.15 meters does _not_ apply here, correct?

I think this could be more clearly worded as
> The restrictions defined for <<Scoring Goals, scoring goals>>, including the ball height limit of 0.15 meters, do not apply here.

MathewMacDougall (changes requested):

chapters/refereecommands.adoc
g3force: yeah, that sounds more clearly, thx.

chapters/refereecommands.adoc
```diff
@@ -128,18 +128,34 @@ It can also be used to resume the game when the game had to be stopped and no te
...
+.. One attacking robot is allowed to approach the ball but not touch the ball
+.. All other robots have to be 1m behind the ball such that they do not interfere the penalty kick procedure at any time.
+. When the <<Normal Start, normal start>> command is issued, the attacker is allowed to <<Ball Manipulation, manipulate the ball>>. The ball has to only move towards the opponent goal, as measured by its x coordinate in the coordinate system of <<Vision, SSL-Vision>>.
+. When the ball is <<Ball In And Out Of Play, in play>>, the defending keeper may move freely again
 
-When the <<Normal Start, normal start>> command is issued, the kicker is allowed to shoot the ball. A goal may be scored directly from the penalty kick.
+A goal is awarded if:
+
+* the ball touches the inner surface of a goal wall or the ground of the goal of the defending team in less than 10 seconds, starting from when the <<Normal Start, normal start>> command is issued
+* the defending team violates any rule
```
tobiasheineken, line 144: What rules might be violated by the defending / attacking team?

chapters/refereecommands.adoc
```diff
@@ -128,18 +128,34 @@ It can also be used to resume the game when the game had to be stopped and no te
...
+The procedure of a penalty kick is as follows:
 
-When the penalty command is issued, one attacking robot is allowed to approach but not touch the ball. This robot will be referred to as the kicker. The defending keeper has to touch the goal line. All other robots have to move behind a line parallel to the goal line and 0.4 meters behind the penalty mark. When these constraints are met, the referee may continue with a <<Normal Start, normal start>> command.
+. The ball is placed by the human referee on the <<Penalty Mark, penalty mark>>.
+. When the <<Penalty Kick, penalty>> command is issued
+.. The defending keeper has to move to the goal line and keep touching it
+.. One attacking robot is allowed to approach the ball but not touch the ball
+.. All other robots have to be 1m behind the ball such that they do not interfere the penalty kick procedure at any time.
+. When the <<Normal Start, normal start>> command is issued, the attacker is allowed to <<Ball Manipulation, manipulate the ball>>. The ball has to only move towards the opponent goal, as measured by its x coordinate in the coordinate system of <<Vision, SSL-Vision>>.
```
tobiasheineken, line 138: I'm pretty sure that even after this change you may not use multiple attacking robots for this penalty. But I cannot see it stated clearly in this PR.

g3force: > All other robots have to be 1m behind the ball such that they do not interfere the penalty kick procedure at any time.

Is this not clear enough?

chapters/refereecommands.adoc
g3force: I think I just copied the rules from one place to another, but still a valid question.

Possible fouls would be: Crashing, dropping parts, dribbling, ball speed, attacker touching keeper in defense area, unsporting behavior.

chapters/refereecommands.adoc
tobiasheineken: I misread `at any time`, just imagining `at the start`. I know where this confusion starts: The text reads like a "first, second, third" where `all other robots` are in the preparation phase. But you're right, the rules still contain that section. It's just me i guess.

PR #15, opened 2020-02-01
Multiple yellow cards turn into red cards instead of penalty kicks

g3force:
Related to #14 

Multiple yellow cards will turn into red cards instead of penalty kicks.


tobiasheineken:
Any thoughts on the fact that this seems a lot less impactful than the old penalty, especially in 2020 where most likely a lot of teams won't be able to start with 11 robots and therefore ignore the yellow (and red) cards entirely?

g3force:
What makes you think that teams will attend with less than 11 robots? When we upgraded to 8 robots, teams played with 8 as well, as far as I remember.

Personally, I found the penalty kicks quite a hard penalty. While this is a good motivation to focus on playing fair, I would prefer to not decide the games on rule confirmatory.

tobiasheineken:
Maybe I'm wrong about starting with less robots. I can't really speak for other teams i guess.

To rephrase the point: Yellow cards have been very impactful when playing with 5 / 6 robots. 
They got a lot less impactful when transitioning to 8 robots, as the relative cost went down.
Also, the introduction of a penalty after 3 yellow cards was to compensate for the fact that at the end of the game it did not matter nearly as much as at the start to get a yellow card: You'd only have to play the remaining time - which is less than the full 2 minutes - one robot down.

That's the reasoning why a penalty was introduced.
This PR does not tackle either of these problems. Playing 10 vs 11 will feel a lot less worse than 5 vs 6 or 4 vs 5. You need 3 yellow cards simultaneously, so playing 8 vs 11 or 7 vs 11 will be felt - but for reasons I don't know yet even 5 vs 8 was kind of OK last year, so 7 vs 11 will be aswell (?)
Also, if the game is close to finish, a red card and a yellow card simply do not have any difference at all. So from that POV, this change is neglect able.

One protocol suggested to use the (new) penalty procedure instead. This makes it that 3 yellow cards still is something to fear, even if the game is close to the end. It also tackles @g3force concern, that the old rules overemphasized rule confirmatory. While I personally don't think so, I understand it and know it's a shared opinion in the league. Therefore, reducing the impact of the penalty sounds like a fair path to me.

g3force:
One of the motivations of this change were to reduce the number of game interruptions. There were also the idea of moving all the penalty kicks to the end of the half. That way, the interruption is avoided and if there are multiple penalty kicks, they could be performed more efficiently after each other.
Would that be an option?

mickmack1213:
As far as I know, postponing penalty kicks to the end of the half is quite unusual, especially in "human" sports.

I know the argument "But in the real soccer it's like..." shouldn't be used, but it's not just soccer it's every main sports (as far as I know) and the league is already very complicated and hard to watch.

I think this change will mainly cause a lot of confusion for the viewers, because of the small spreading of postponing penalty kicks. And I believe more understandable games are even more important than quicker games.

MathewMacDougall:
I also don't totally like postponing the penalty kicks to the end of the half / game, because I also think it would cause lots of games to feel like they're being decided by a penalty shoot-out. Even though a penalty shoot-out is less one-sided than a penalty shot, making it potentially the last thing that happens in a game would definitely make it confusing for viewers and make it feel like it is still deciding the game, rather than allowing for more of a "comeback" scenario during normal gameplay.

I know "feels like" arguments aren't very good, just my thoughts.

andre-ryll:
I think the main reason for this change was to reduce interruptions and to overall have shorter matches. So regardless of the severity of a penalty or one-on-one shootout I think we all agree that a red card is simply quicker and hence best fixes the issue this PR is about.

tobiasheineken:
@andre-ryll 

While the PR fixes this stated issue, we still have to discuss its overall impact.
For example: Let's assume that we all agree that currently there are not enough goals in the game. (A reasonable person might think so, games with more goal are normally less luck based and more of a skill based game.)
To do that, we could simply remove the goalie and enforce that at maximum half of your robots may be in your half of the filed. After these changes, there will be more goals.
I think we all agree that this would not be a good rule even in the imagined scenario.
So even if we want to get shorter matches, we have to be careful when doing so. Simply stating that "it will decrease the time for a match" is not sufficient.

 As of #14, we don't have the old penalty kick anymore. Also, @g3force commented that the current rules have a (maybe too strong) focus on simply obeying the rules which might not be the thing that we want. @mickmack1213 and @MathewMacDougall commented that postponing the (weaker) one-on-one penalty doesn't sound appealing to them.
Accepting this PR would rollback the idea behind the foul-counter and this yellow-card into Penalty rule: punishing a team for rule violations (especially collisions) even if there is very little game time left. I think the decrease of violent collisions can be attributed to that change and is good thing to the league.

This PR does not handle yellow cards late in the game: If there is very little game time left, there's no difference in a red card or a yellow card, and even a yellow card is weaker than it should be.

I do understand that we don't want to stop the game right away for the third yellow card penalty one-on-one. I suggest to keep the penalty (which is the new one after #14), and do that procedure as soon as the game is already interrupted. It might not be the best solution for games with walls, but walls do come with there own set of problems and let's solve one at a time.

We won't really safe time with this change. We even have to discuss if the one-on-one time should be counted as game time (I tend towards no, but I'm unsure). But we do safe time with other changes, and keeping the "late game yellow card" penalty seems more important to me.
It solved the issue state by me, while covering the concerns by g3force, mickmack and Mathew (at least I hope)

MathewMacDougall:
I agree with @tobiasheineken but just want to clarify: I am not against the 1-on-1 penalty shootout, I am only opposed to them being delayed until the end of the game, as I think it ruins some of the gameplay and spectator experience.

mickmack1213:
@tobiasheineken I don't really think that's a good solution. With all the other rule changes, 1 or 2 minutes of uninterrupted gameplay won't be that uncommon. Especially with walls. It can happen a lot within 2 minutes. So situations like: "The foul happend over 1 min before the Penalty execution" or "More than 1 Penalty needs to be applied" won't be uncommon as well.

I think postponing the execution to the stop is only a minor improvement compared to postponing it to the end of half/game. And I really want to speak up for keeping the time between the foul and the penalty execution as short as possible, at best instant. It's just a lot easier to understand while watching and i believe it's also simpler in the rulebook and the game controller.

I don't really have an opinion on what kind of penalty is applied, Red card / Free kick / 1on1 / e.g. But please consider to execute the penalty as short as possible after the foul.

g3force:
I'm a bit out now, there are quite some suggestions. Lets make a quick poll to see if we need to discuss more on this:

https://poll.lab.io/E7qP-EpNgtIEu2ldF04MXg

Penalty kicks refer to the one-on-one kicks.

g3force:
/polls "Keep the PR as is" "Perform penalty kick immediately" "Penalty kicks on the next stop" "Penalty kicks on the end of the game half"

g3force:
Anonymous poll result:

* Keep PR as is: 4
* Use penalty kick: 3
* Penalty kick on next stop: 1
* Penalty on end of half: 0

The delayed options are out. It remains to be decided to use the penalty kick or the red card.

We should also consider that if we decide for the penalty kick, there are probably a lot more of them compared to last year (where there were none actually) and teams might not have prepared for it yet. We could start with the red card and reconsider the penalty kick after the RC. Then, teams have more time to prepare for the new penalty kicks.

tobiasheineken:
A red card **does not solve the issue** why this rule was introduced a few years back.

If we don't want to push teams towards implementing the new penalty, we could drop this rule completely and do not act in any shape or form if a team gets multiple yellow cards.

g3force:
Dropping the multiple foul rule would indeed by another option if we find the yellow cards sufficient.

What issue do think of, when you say it does not solve the issue?

tobiasheineken:
> Accepting this PR would rollback the idea behind the foul-counter and this yellow-card into Penalty rule: punishing a team for rule violations (especially collisions) even if there is very little game time left.

>  This PR does not handle yellow cards late in the game: If there is very little game time left, there's no difference in a red card or a yellow card, and even a yellow card is weaker than it should be.

g3force:
thx. So, if the main motivation of this rule is to adequately punish teams late in the game, the rule shot over its goal. There were more fouls happening than we had expected.

In the mean time, it would prefer dropping the rule completely to simplify things again.

Btw: If we keep the penalty, we need to think about how and when to stop the game, because we do not stop for yellow cards anymore, so the game would still be running. That's one more special case again...

tobiasheineken:
The main motivation was to introduce something to force teams to obey the rules, even the smaller ones, no matter if the game is already close to the end.

I still agree with its concept a lot. First, a yellow card is way less impactful when playing 8 vs 8 (or even 11 vs 11) compared to 5 vs 5 or 6 vs 6.
Second, as the game comes to an end, there's really no reason not to go for a more aggressive stance. Regular soccer solves this issue by blocking this particular player for another full match, which doesn't work in our sport. So we went the basketball way: Add something that's almost certainly a goal.

How about not changing anything regarding this issue and stick to the (already implemented) penalty with a mark to intend to switch to the shootout-variant soon?

g3force:
You have implemented the penalty mark kick, but other teams may not. I think it is a bad idea to keep the penalty mark kick in the rules, if it will definitely be removed next year. There is no motivation in implementing or improving it for the teams anymore...

In the mean time, the procedure for the new penalty kick is clarified in #32, so that would not be an issue anymore.

Still, I think the penalty kick is a too harsh penalty for obeying the rules. I have the impression that teams generally don't collect yellow cards by intention and are working on reducing the fouls. And if individual teams don't care, that might again be a case of unsporting behavior, in which case the team could be warned, before taking any more serious action.

So, after all, I'm still in favor of dropping the Multiple Yellow Card rule completely.

g3force:
Actually, after thinking about it a bit more, I'd like to keep the rule as is:

The original idea of the rule change as it is now was: Punish teams that perform a lot of fouls in a high frequency harder and more persistent using red cards.

tobiasheineken:
I think this is a mistake. The only reason why I cared about yellow cards and in extension the foul counter was the almost certain goal via penalty kick. Playing 6 vs 8 didn't feel so bad last year and three yellow cards _at once_ was never a concern to us.
If this goes into the rules, I will stop trying to avoid crashing, which is stated as an open problem in our ETDP, as the main concern switches from the penalty to "loosing possession". Together with #32, where crashing does not affect anything regarding possession, as the game is not stopped, there is no reason to think about these things anymore.

It is really inconsistent to push towards better collision avoidance in 2018 and 2019 and dropping it almost 100% in 2020. Maybe I'm wrong and one team will show me to fear a yellow card, but I'm not convinced yet.

And don't start stating "unsporting behavior" as a reason here. I do not want collisions. Most likely we won't increase our collisions by a lot this year. I just cannot justify allocating resources to solve or work on this issue if the expected results are minor at best.

tobiasheineken:
I have one final compromise:

How about an option to each team to use the old or the new penalty kick for this year? Like in Japan, where teams where encouraged to play with ball placement, but didn't need to if they didn't want to? Teams could simply decide before the game and in any of their timeouts / halftime to switch between the old or the new penalty kick.

If we do it this way, teams that don't want to improve their old penalty kick can already work on the new one, and teams who don't find the time for a good implementation can fallback to the old one. There are not that many teams out there without an implementation for the old penalty kick. This argument is very weak.

g3force:
What about this: When you committed say 10 collisions, you get a warning to reduce your robot speed. If you are still committing collisions, say another 5, you get a penalty kick or get disqualified due to unsporting behavior (both penalties are already allowed for this), based on how frequent and how hard you collided with others.

Does this sound motivating enough for you? ;)

Collisions could only reliably be detected since we use the autoRefs. Before, human refs had too few data to decide if a team actually performed too many collision. But now, the data is there, so the human ref has a better decision base.

PR #17, opened 2020-02-02
Change lack of progress timeout and behavior

rhololkeolke:
Play now continues with a Force Start instead of an Indirect Free Kick.

Division A timeout changed to 5 seconds. Division B timeout remains at 10 seconds.

See proposed rule change spreadsheet: https://docs.google.com/spreadsheets/d/1LfNNrE4ld0IkRGXSkKSKOnPDipdKbuH-InZJY2kATs4/edit?usp=sharing

### DIFF
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -60,8 +60,9 @@ Chapter <<Offenses>>, section <<Minor Offenses>>:
 | Event | Applicability | Command | AutoRef
 
 | <<Aimless Kick [small]#(_division B only_)#, Aimless Kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
-| <<Lack Of Progress>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Force Start>> | icon:check[role="green"]
+| <<Lack Of Progress>> during <<Kick-Off>>, <<Direct Free Kick>>, <<Indirect Free Kick>> | <<Ball In And Out Of Play, ball in play>> | Ball is considered in-play and manipulatable by both teams | icon:check[role="green"]
 | <<Lack Of Progress>> ball in <<Defense Area, defense area>> | <<Ball In And Out Of Play, ball in play>> | <<Indirect Free Kick, indirect free kick>> | icon:check[role="green"]
+| <<Lack Of Progress>>, otherwise | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Force Start>> | icon:check[role="green"]
 | <<Double Touch>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
 | <<Attacker In Defense Area>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
 | <<Attacker Touches Robot In Opponent Defense Area>> skipped | <<Ball In And Out Of Play, ball in play>> | no command | icon:check[role="green"] (<<Advantage Rule>>)
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -60,7 +60,8 @@ Chapter <<Offenses>>, section <<Minor Offenses>>:
 | Event | Applicability | Command | AutoRef
 
 | <<Aimless Kick [small]#(_division B only_)#, Aimless Kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
-| <<Lack Of Progress>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
+| <<Lack Of Progress>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Force Start>> | icon:check[role="green"]
+| <<Lack Of Progress>> ball in <<Defense Area, defense area>> | <<Ball In And Out Of Play, ball in play>> | <<Indirect Free Kick, indirect free kick>> | icon:check[role="green"]
 | <<Double Touch>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
 | <<Attacker In Defense Area>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
 | <<Attacker Touches Robot In Opponent Defense Area>> skipped | <<Ball In And Out Of Play, ball in play>> | no command | icon:check[role="green"] (<<Advantage Rule>>)
@@ -113,4 +114,4 @@ This is a complete list of differences between <<Divisions, division>> A and <<D
 * Division A plays with <<Number Of Robots, more robots>> than division B.
 * The automatic <<Ball Placement, ball placement>> procedure is mandatory for division A and optional for division B.
 * The <<Aimless Kick [small]#(_division B only_)#, aimless kick>> rule only applies to division B.
-* There is a smaller time window in division A for taking a free kick before <<Lack Of Progress, lack of progress>> is called.
+* There is a smaller time window in division A before <<Lack Of Progress, lack of progress>> is called.
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -44,7 +44,7 @@ kick>>, <<Indirect Free Kick, indirect free kick>>, after the time
 expires the ball is considered <<Ball In And Out Of Play, in play>>
 and both teams may manipulate the ball.
 
-In all other cases a <<Stop>> is issued followed by a <<Force Start>>
+In all other cases a <<Stop, stop>> is issued followed by a <<Force Start, forced start>>.
 This includes situations where both teams can manipulate the ball, but
 the referee determines that no progress is being made.
 
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -12,13 +12,34 @@ All minor offenses are listed below.
 A kick is aimless when after the ball touched a robot, it subsequently crossed the midline and then its opponent's goal line outside the goal without touching another robot. A kick-off kick cannot be aimless, as the ball is located at the midline and does therefore not cross it.
 
 ==== Lack Of Progress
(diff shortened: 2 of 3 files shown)


chapters/refereecommands.adoc
```diff
@@ -122,7 +122,7 @@ Indirect free kicks are used to restart the game after a <<Minor Offenses, minor
 When the force start command is issued, the game is immediately resumed and both teams are allowed to approach and <<Ball Manipulation, manipulate the ball>> again.
 
 .Usage
-The referee can issue a stop command followed by force start if there is a clear lack of progress for at least 10 seconds while both teams are allowed to approach and <<Ball Manipulation, manipulate the ball>>.
+The referee can issue a stop command followed by force start if there is a clear lack of progress for at least 5 seconds while both teams are allowed to approach and <<Ball Manipulation, manipulate the ball>>.
```
tobiasheineken: This fails to mention Div B / Div A differences

chapters/offenses.adoc
```diff
@@ -12,9 +12,9 @@ All minor offenses are listed below.
 A kick is aimless when after the ball touched a robot, it subsequently crossed the midline and then its opponent's goal line outside the goal without touching another robot. A kick-off kick cannot be aimless, as the ball is located at the midline and does therefore not cross it.
 
 ==== Lack Of Progress
-There is a lack of progress if only one team is allowed to <<Ball Manipulation, manipulate the ball>> (<<Kick-Off, kick-off>>, <<Direct Free Kick, direct free kick>>, <<Indirect Free Kick, indirect free kick>>, <<Penalty Kick, penalty kick>>) and does not bring the ball <<Ball In And Out Of Play, into play>> in time. The time limit is 5 seconds for <<Indirect Free Kick, indirect free kicks>> and <<Direct Free Kick, direct free kicks>> in division A and 10 seconds in all other cases.
+There is a lack of progress if only one team is allowed to <<Ball Manipulation, manipulate the ball>> (<<Kick-Off, kick-off>>, <<Direct Free Kick, direct free kick>>, <<Indirect Free Kick, indirect free kick>>, <<Penalty Kick, penalty kick>>) and does not bring the ball <<Ball In And Out Of Play, into play>> in time. The time limit is 5 seconds in division A and 10 seconds in division B.
```
tobiasheineken: The additional time in kickoff has been very important last year. As Kickoffs are pretty rare, don't allow preparation during ball placement, and require a lot of driving. Also, the same issue might arise in the (new / old) penalty kick, where you're not able to prepare during ballplacement.

chapters/appendix.adoc
```diff
@@ -60,7 +60,7 @@ Chapter <<Offenses>>, section <<Minor Offenses>>:
 | Event | Applicability | Command | AutoRef
 
 | <<Aimless Kick [small]#(_division B only_)#, Aimless Kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
-| <<Lack Of Progress>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Indirect Free Kick>> | icon:check[role="green"]
+| <<Lack Of Progress>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Force Start>> | icon:check[role="green"]
```
tobiasheineken: Force Start doesn't work if the ball is inside the defense Area.

tobiasheineken (changes requested):

chapters/offenses.adoc
g3force: Good catch. I think, the intention was to reduce the time for free kicks only. Kick off and especially the one-on-one penalty kick should not be changed here.

chapters/appendix.adoc
g3force: just like free kick. But as the game is stopped before the next command, there will also be a ball placement, if the ball is not at valid position.

tobiasheineken: It is well-defined what happens if a free-kick is given at an invalid position (as far as i remember).

I do not remember an analogous rule regarding force start. I might be wrong, though, but that's the reason I mentioned it here.

tobiasheineken: I've found it. It's referenced in this document (see xxx for details regarding the positon), and is explained here:
```

Definition

The ball placement position for a free kick depends on the event that led to the free kick. This position is valid if there is at least 0.2 meters distance to all field lines and 1 meter distance to either defense area. If an event requires the ball to be placed at a position that contravenes this rule, it has to be placed at the closest valid position instead.
```

There is no rule how to handle the same situation for a force start.
If there's a ball placement into a valid position, we should note that here.
Also, is ballplacement into force-start already a valid transition? Because it feels to me like you'd have a huge benefit if you're getting to place the ball as you might be much closer than your opponent after the ballplacement.

If you have to drive away it should be faster to resume the game via an indirect free kick than via force start. Sadly, we dumped Indirect free kicks so now we're lost as a direct free kick is a much stronger tool in this situation :(

g3force: as far as I read the rules, there is no specific definition for how the ball is placed for free kicks. It just says, the ball must be placed after stoppage and there is the condition with the penalty area.

g3force: oh, you were faster. And I actually misread the rules. I thought, the distance to the defense area is defined in the ball placement section.

g3force: oh wait, but lack of progress can only happen from free kick, kick off and penalty kick positions. With the old penalty kick, there might be an issue with the position, but with the new procedure, all positions should already be fine.
Additionally, the whole idea of the force start was actually to not stop the game in between. Otherwise, we would not really save time with the force start.

tobiasheineken: This whole discussion is centered around the lack of progress inside the defense area, if the keeper is unable to clear the ball in 5 seconds...

g3force: @rhololkeolke I think that still needs to be specified somewhere. The force start is only mentioned in the appendix and the usual procedure is to first stop the game.

g3force: oh, now I see. It is an additional sentence in the rules. Would be better to attach the comment directly to this line. I'll do it.

chapters/offenses.adoc
```diff
@@ -12,9 +12,9 @@ All minor offenses are listed below.
...
 
-There is also a lack of progress if the ball is inside a team's <<Defense Area, defense area>> for 10 seconds, since the keeper is the only robot that is allowed to <<Ball Manipulation, manipulate the ball>>.
+There is also a lack of progress if the ball is inside a team's <<Defense Area, defense area>> for 5 seconds, since the keeper is the only robot that is allowed to <<Ball Manipulation, manipulate the ball>>.
```
g3force: As @tobiasheineken pointed out, for this situation, a force start is not reasonable.

The autoRefs separate this already: There is a `KeeperHeldBall` and a `KickTimeout` game event. I think we should separate this in the rules as well and make it two different minor offenses.

tobiasheineken: Still, we've lost the Indirect Freekick for KeeperHeldBall which makes me sad.

g3force: You can create a new discussion in the respective PR. If there is enough evidence, that an indirect free kick is still reasonable (like in this situation), what about merging all free kicks to indirect free kicks? Would this solve the issue, or is there actually something else that makes you uncomfortable with this rule change?

chapters/appendix.adoc
tobiasheineken: Stop, then Force Start implies that both teams have to keep distance before force-start can be given. If this rule is not applied here, that should be worded differently. @g3force did I understand you correctly that the improved speed should be achieved by not stopping the game? If so, it might be an option to remove "lack of progress" in the case of freekicks and instead state in "ball goes into play" that the ball goes into play at latest 5 seconds after the Indirect / Direct command was issued. This results in slightly different rules (Force-Start is given vs. teams have to count and wait on their own), but might be the more straight-forward way of implementing this change.

g3force: I really like this idea. The teams do not even need to count themselves, because the referee messages already contain the counter since 2019. The only disadvantage is, that teams actually have to modify their system slightly, but that should be fine, as the change is small and not mandatory to confirm to the rules.

mickmack1213:
What's about double touch? With this implementation you can avoid a double touch foul of by just waiting 5s and dribble the ball afterwards. The kicking team is much closer to the ball and has a clear advantage. 
So the double touch rule must be enforced after the 5 seconds as well.

g3force:
The rules state:
> When the ball is brought into play following a kick-off, direct free kick, indirect free kick or penalty kick, the kicker is not allowed to touch the ball until it has been touched by another robot or the game has been stopped.

So, with the suggestion from @tobiasheineken the double touch rule is effective after the 5 seconds as well. When directly sending a force start without a stop, it would also keep effective.

But stop, then force start would then definitely require a small pause so that both teams have the same distance to the ball. This variant is out, I guess...

rhololkeolke:
I rewrote the lack of progress section. There is a table with timings for different situations and divisions. It is all 10s for Div B. Kick-off and Penalty are still 10s for Div A. All other situations are 5s for division A.

For the case where the ball is in the defense area I've kept the indirect free kick procedure. Depending on the outcome of #3  we may need to reevaluate this procedure. My alternative suggestion is ball placement outside the defense area, then stop, then force start. But we would need to decide where the placement should go in that case.

In all other situations, the procedure has been changed to stop then force start.

g3force:
I really liked the idea of @tobiasheineken to continue the game without any stop and force start. The stop and force start would not save any time compared to the current rules.

rhololkeolke:
@tobiasheineken, @g3force Does that address the suggested change?

Kick-off and free-kicks auto continue. Manual calls do a stop and force start. When the ball is only reachable by a single team we use the free-kick rules.

chapters/offenses.adoc
```diff
@@ -12,13 +12,41 @@ All minor offenses are listed below.
...
+If there is a lack of progress due to the ball being inside a team's
+<<Defense Area, defense area>> past the timeout, then a <<Indirect Free
+Kick, indirect free kick>> is issued for the attacking team.
+
+In the case of <<Kick-Off, kick-off>>, <<Direct Free Kick, direct free
+kick>>, <<Indirect Free Kick, indirect free kick>>, after the time
+expires the ball is considered <<Ball In And Out Of Play, in play>>
+and both teams may manipulate the ball.
+
+In all other cases a <<Stop>> is issued followed by a <<Force Start>>
```
```suggestion
In all other cases a <<Stop, stop>> is issued followed by a <<Force Start, forced start>>.
```

chapters/refereecommands.adoc
```diff
@@ -122,7 +122,8 @@ Indirect free kicks are used to restart the game after a <<Minor Offenses, minor
 When the force start command is issued, the game is immediately resumed and both teams are allowed to approach and <<Ball Manipulation, manipulate the ball>> again.
 
 .Usage
-The referee can issue a stop command followed by force start if there is a clear lack of progress for at least 10 seconds while both teams are allowed to approach and <<Ball Manipulation, manipulate the ball>>.
+The referee can issue a stop command followed by force start if there
```
g3force:
```suggestion
The referee can issue a <<Stop, stop>> command followed by <<Force Start, forced start>> if there
```

g3force (changes requested):

rhololkeolke:
Closed in favor of #32

PR #18, opened 2020-02-02
Change Attacker touches robot in defense penalties

rhololkeolke:
Remove stoppage and free kick.

Goals within 2 seconds of end of penalty do not count. Foul counter increases.

Rephrased Minor Offenses section as the old text is no longer accurate.

### DIFF
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -74,7 +74,7 @@ Chapter <<Offenses>>, section <<Fouls>>:
 | Event | Applicability | Command | AutoRef
 
 | Multiple <<Fouls>> | <<Ball In And Out Of Play, ball out of play>> | <<Yellow Card>> | icon:times[role="red"] (handled by the game controller)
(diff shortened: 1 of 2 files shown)


tobiasheineken:
I dislike that in this change (and similar ones) the games keeps running and goals are not awarded, what really makes it hard to follow the game. Also, seconds where the clock counts down while it's in the best interest of at least one teams to not let the game progress seem bad for the game.

g3force:
I find the game harder to follow, when it is stopped and I also think that such a counter is reasonable or at least I do not know of any better alternative. 
I also don't think it is bad for the game, if a team can not score a goal. 10s is not that much time, so it can either just keep going or just try to keep ball control, which can also be interesting.

We will probably add the count down to the status board as well.

MathewMacDougall:
I'm fine with this change because it already happens pretty infrequently, and we don't really know how these periods of time where goals are not allowed will affect the spectator experience. Personally I'm willing to try it.

rhololkeolke:
Closed in favor of #32

PR #19, opened 2020-02-02
Remove stoppage after crashing

rhololkeolke:
Game continues. Foul counter is increased.

Rewrote Fouls section.

### DIFF
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -41,11 +41,29 @@ NOTE: Dribblers can still be used to dribble large distances with the ball as lo
 A robot must not accelerate the ball faster than 6.5 meters per second in 3D space.
 
 === Fouls
-Fouls are violations of the rules that result in a <<Direct Free Kick, direct free kick>> for the opposing team. The free kick will be shot from the position where the offense began happening (see <<Direct Free Kick, the direct free kick rules>> for the exact ball position rules). If the foul happened while the ball is <<Ball In And Out Of Play, out of play>>, no free kick is given.
 
-Every third foul of the same team results in a <<Yellow Card, yellow card>>.
+Fouls are violations of the rules that result in a some combination of
+the following penalties:
 
-In case of severe fouls, the referee can also issue a <<Yellow Card, yellow card>> or a <<Red Card, red card>>.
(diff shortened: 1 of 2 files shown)


chapters/offenses.adoc
```diff
@@ -68,7 +86,9 @@ In this case, the placement procedure is restarted.
 NOTE: This rule does not cover all cases of ball placement interference. The <<Referee, referee>> is encouraged to call fouls if the non-placing team is obviously interfering with the ball placement.
 
 ==== Crashing
-At the moment of collision of two robots of different teams, the difference of the speed vectors of both robots is taken and projected onto the line that is defined by the position of both robots. If the length of this projection is greater than 1.5 meters per second, the faster robot committed a foul. If the absolute robot speed difference is less than 0.3 meters per second, both conduct a foul but the game will not be stopped.
+At the moment of collision of two robots of different teams, the difference of the speed vectors of both robots is taken and projected onto the line that is defined by the position of both robots. If the length of this projection is greater than 1.5 meters per second, the faster robot committed a foul. If the absolute robot speed difference is less than 0.3 meters per second, both conduct a foul.
+
+The game does not stop.
```
tobiasheineken, line 91: So an unfair crash where one party is clearly in fault into a goal is a valid move?

g3force: I do not quite get what you mean, but if a team does something unfair, this can usually be punished with unsporting behavior.

tobiasheineken: I'd assume that there's a grace period after each foul that did not stop the game where you cannot score a goal. I mentioned in another PR, why I don't like this concept, but you seem to be committed to it.

But this foul does not come with its grace period. Which makes it confusing to me - why not solve it at a general location for all fouls? Is a "hard tackle" just before the goal is shot now ok? Do I have a easy way to decide if that foul contains a grace period or not?

g3force:
As mentioned in #29, the advantage rule is obsolete with this change.

chapters/offenses.adoc
g3force: I agree that we should define the grace period at a central place to keep it simple and intuitive.

rhololkeolke:
Closed in favor of #32

PR #20, opened 2020-02-02
Ball placement interference

rhololkeolke:
Don't reset. Increase timer by 10 seconds.


tobiasheineken:
IMO this PR (and many other that share the prologue in foul) has a huge flaw:

Previously, committing a foul was a valid description. Committing a foul was a well-defined thing, easy to understand and implement on your own. Now, committing a foul is not well-defined, as the PR states itself in ` See below for details on the specifics of the penalties for each offense type.` This requires to add additional info to each foul. Which is not done for this foul, nor for any other foul.

While I do agree that ball placement interference was triggering way to often last year, the whole prologue is not needed for this PR and might be a bad idea in general.

We have been working very hard in the previous years to increase understandably and simplicity regarding the rules and undoing or degrading this hard work seems counter productive.

rhololkeolke:
Closed in favor of #32

PR #21, opened 2020-02-02
Change penalties for Defender Too Close to Ball

rhololkeolke:
Game continues. Foul counter increases.

### DIFF
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -114,7 +114,8 @@ NOTE: Since the stop command is used for manual ball placement and <<Robot Subst
 
 ==== Defender Too Close To Ball
 A robot's distance to the ball must be at least 0.5 meters during an opponent <<Kick-Off, kick-off>>, <<Direct Free Kick, direct free kick>> or <<Indirect Free Kick, indirect free kick>>.
-The game is resumed with the same command that was issued before the foul.
+
+Game continues but <<Fouls, foul counter increases>>.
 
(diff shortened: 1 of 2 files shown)


chapters/offenses.adoc
```diff
@@ -96,7 +114,8 @@ NOTE: Since the stop command is used for manual ball placement and <<Robot Subst
 
 ==== Defender Too Close To Ball
 A robot's distance to the ball must be at least 0.5 meters during an opponent <<Kick-Off, kick-off>>, <<Direct Free Kick, direct free kick>> or <<Indirect Free Kick, indirect free kick>>.
-The game is resumed with the same command that was issued before the foul.
+
+Game continues but <<Fouls, foul counter increases>>.
```
MathewMacDougall, line 118: Is there a cooldown period between increasing the foul counter? If a defending robot just stayed too close to the ball, how fast would it accrue fouls?

MathewMacDougall (changes requested):

chapters/offenses.adoc
g3force: At a minimum, the foul counter should only be increased once per robot and free kick. As long as the violating robot is just a bit too close, that should be fine.

It starts getting interesting, when the defender actually drives towards the ball and hinders the attacker to perform the free kick. That's not unusual if the ball is not detected well and a team thinks it can approach the ball. I think, in the past good referees punished this with unsporting behavior when it happened too frequently.
Are we fine with keeping this to the referee?

MathewMacDougall: Fine with me, I just mostly was curious about mild infractions.

tobiasheineken: I do not agree that we can rely on the referee. The whole `hindering the attacker` point is the whole reason why we resumed the game with the same command, resetting the state for a decent team to a clear stop-state and retry again.

rhololkeolke:
Closed in favor of #32

PR #23, opened 2020-02-02
Ball speed penalties

rhololkeolke:
No stoppage.

Goals with kicks exceeding limit do not count. Increase the foul counter.

### DIFF
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -48,6 +48,8 @@ NOTE: Dribblers can still be used to dribble large distances with the ball as lo
 ==== Ball Speed
 A robot must not accelerate the ball faster than 6.5 meters per second in 3D space.
 
+Violations do not stop game play. Goals scored with kicks exceeding the ball speed limit will not be counted. Increase <<Fouls, foul counter>> by one.
+
 === Fouls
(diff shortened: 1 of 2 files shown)


chapters/offenses.adoc
```diff
@@ -40,6 +48,8 @@ NOTE: Dribblers can still be used to dribble large distances with the ball as lo
 ==== Ball Speed
 A robot must not accelerate the ball faster than 6.5 meters per second in 3D space.
 
+Violations do not stop game play. Goals scored with kicks exceeding the ball speed limit will not be counted. Increase <<Fouls, foul counter>> by one.
```
MathewMacDougall, line 51: How is gameplay restarted after a goal is not counted?

MathewMacDougall (changes requested):

chapters/offenses.adoc
g3force: Good point. It is defined in the "Scoring goals" sections for other invalid goals: It is counted as if the ball crossed the goal line outside of the goal. A reference to this sections should probably be sufficient, but I haven't checked the formulations in detail.

g3force, line 51: We should use the grace period here as well. Making a fast pass and than scoring a goal should also not be allowed.

rhololkeolke:
Closed in favor of #32

PR #27, opened 2020-02-03
Challenge flag (Both Div)

leonardoSCosta:
From the rule change proposal:

> In American football there is a challenge flag: A team can challenge a decision of the referee
> If referees decision was correct, team loses a timeout
> If referees decision was incorrect, correct decision is applied
> Only one ruling may be challenged at a time.
> 
> We consider giving each team about two flags. The exact number is yet to be determined.
> 
> The rationale is that we are losing a lot of time discussing referee decisions. It will make teams think twice before discussing a situation.


chapters/challengeflag.adoc
```diff
@@ -0,0 +1,12 @@
+== Challenge Flag
+
+In American football there is a challenge flag. A team can challenge a decision of the referee:
```
g3force: I would move the american football part to a note (sth. like: This is inspired by challenge flags in american football).

g3force (changes requested):

chapters/challengeflag.adoc
```diff
@@ -0,0 +1,14 @@
+== Challenge Flag
+
+A challenge flag allows teams to challenge a decision of the referee:
+
+. If referees decision was correct, team loses a timeout.
+. If referees decision was incorrect, the correct decision is applied and the team doesn't lose a timeout.
+
+NOTE: In both cases the team loses one flag.
```
tobiasheineken: Why do you loose a flag if you're wrong?
As usual, quoting a sport (especially if it is non-soccer) does not qualify for an explanation.


I'm perfectly fine with loosing the flag if you're wrong. (I still do not really understand why you also loose a timeout but what ever...)
But while this rule does encourage better understanding of the rules, which is a really good thing, loosing your challenge-flag when you're correct reduces the encouragement to know the rules by a lot.

g3force: Given that we took the rule from another sport, the most obvious definition is the one from the other sport, until there are good reasons to do it differently.

What we are currently missing from the NFL is, that a team gets a third challenge, if both challenges were correct. I have no strong opinion whether we add this or not, it just adds a bit more of complexity, if we actually implement this in the GC. I would also be fine with giving the challenge flag back, if it was successful.

If you would not loose a timeout, there would be no high risk in requesting the challenge, especially later in the game. I find this quite reasonable and it seems to work well in the NFL.

chapters/challengeflag.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+Only one ruling may be challenged at a time.
+
+Each team will receive two flags at the start of the game.
```
g3force: It should be added that a challenge can only be requested, if the team has at least one timeout left.

mickmack1213:
I agree with @tobiasheineken. It might create some weird situations if you can't regain your challenge flag. You might have to to decide between using the flag and speculating to save it for a later situation, where it might be even more important/valuable.
And I think to keep the complexity of the GC and rules as low as possible it's a good idea to just keep the flag if you're correct.

With these rules, a punishment for using the flag wrongly is even more important. If it's for free you have one "free shot" to delay the game and waste time without any risk.

I think that is a nice combination, which can be tested and evaluated during RoboCup.

chapters/challengeflag.adoc
rhololkeolke: At the moment, we are still allowing the refs to review logged data in the autoref. This can take a lot of time. Especially since not all refs/operators are super familiar with the autoref interfaces. What we don't want, is for every disagreed call to stop the game and waste time reviewing these logs, discussing with team leaders, and discussing with the refs/assistant refs.The challenge flag puts an upper limit on how many times in a match this can be done.

Even with this limit, we don't want teams to feel they can challenge plays without a strong basis. For example, if they were scored on, we don't want the calculus to be that its worth losing a challenge flag just on the hopes that the ref will change their mind on a play. Hence, the penalty of losing a timeout if the team is wrong.

If we think loss of a timeout is too harsh a penalty then we can adjust what is lost. We can also increase the challenge flag numbers. Or if we don't want a limit on challenges, but want to keep the penalty, we can just say no replays, the ref just gets to hear the case of the challenger and then reaffirm or overrule. Although in the last case, I still worry that will eat up a lot of time.

I think overall, we just want to formalize what and when teams are allowed to contest the ref. It was my impression that there has been a decent amount of ref arguing in previous years and this would put firm rules as to what is allowed.

tobiasheineken: @rhololkeolke I agree on the fact that a formalized way of "interacting with the referee" is better than the current one. I disagree on "we don't want teams to feel free they can challenge plays without a strong basis", as this is - as you stated yourself - just a formalized way what is and what isn't allowed. And changing nothing but reducing the number of false claims by limiting false claims to two per team would already an improvement. Teams don't use all of their timeouts all the time, why would they use all of their Challenge-Flags?

However, I have no problem with these timeout penalties as described in the PR or by @mickmack123. If you think that helps, go ahead.

It's a known fact that we don't have a lot of good referees in the league.
I hope that this PR has an effect of referee-quality, as it's now in the best interest for a team to know rules by heart. Which would be the major reason this PR speeds up games (not the Challenge itself, but the need for less challenges just because the rules are known universally.) 
But limiting these challenges even if you are correct is a setback.
If our rules should be important, we need to be able to complain if they are not enforced correctly.
Loosing that right due to some kind of offense (like being wrong too often) sounds reasonable, loosing it "just because the referee is so bad" doesn't.

I have no clue why the time it takes for such a challenge to resolve should have any impact on this more fundamental discussion. All these things - replays, discussions, assistant - are already in the game. Limiting it and adding a cost if you're wrong already **will** bring them down compared to previous rules, where you basically had infinite of these. Saying something like "you may not use any tools" for a discussion with the referee seems counter-productive: These information are key for a referee to review the situation. If it's not a simple - I thought the rules where, but they aren't - case, these information are crucial. And again, this PR will limit the complains - by one way or another. Don't overshoot the goal.

g3force (approved):

chapters/challengeflag.adoc
g3force: I have increased the number of flags to three as a comprise. That's already three more than teams were allowed to raise last year (strictly speaking).

Let's see how this will work out in practice.

PR #29, opened 2020-02-12
Resume at the location of the ball when play is stopped after foul

g3force:
A rule change proposal discussed in the Open Meeting at last RoboCup and announced in one of the announcement mails was:

> In the event that play is stopped because of a foul (for example due to a bot collision), play will resume at the location of the ball when play is stopped (current rules resume play at the site of the offense).

This PR implements this proposal


tobiasheineken:
Does this make the advantage-rule obsolete?

g3force:
This PR only influences where the game will continue. The advantage rule was established to avoid stopping the gaming during promising situations.

But the advantage rule will be obsolete, if the game is not stopped for crashes and 'attacker touches robot in opponent defense area'.

rhololkeolke:
This seems fine. We just may have to revisit this section again if we end up removing either Indirect Free Kicks (#3) and after the foul changes are accepted and the groupings are revised.

PR #6, opened 2020-01-20
Change Div A num robots to 11

rhololkeolke:
From Rule Change Proposal Doc:

[Number of robots increased to 11 (DivA only)](https://docs.google.com/document/d/1421rBfoks2ojxyCz_Cjwfq3V2HuvZoVErzx2lxaHvbU/edit#heading=h.pvqxa5umfqa9)
Each team is allowed to play with up to 11 robots at a time, instead of 8.

PR #30, opened 2020-02-13
Increase defense area and goal size

leonardoSCosta:
As pointed out by some teams, the increase in the number of robots (#6 ) might create the possibility to block a large area of the defense and goal area. To avoid that, we are increasing both defense area and goal size. The dimensions were initially increased by 50%, but we are open to any suggestions.

### DIFF
--- a/chapters/playingenvironment.adoc
+++ b/chapters/playingenvironment.adoc
@@ -42,7 +42,7 @@ A mid-line runs along the length of the field, passing through the center of the
 The center mark is indicated at the midpoint of the halfway line. A circle with a diameter of 1 meter is marked around it for both divisions.
 
 ===== Defense Area
-A defense area is defined as a rectangle touching the goal lines centrally in front of both <<Goals, goals>>. The size of the defense area is 2.4 meters times 1.2 meters for division A and 2 meters times 1 meter for division B, as shown in figures <<field-dimensions-a, 1>> and <<field-dimensions-b, 2>> respectively.
+A defense area is defined as a rectangle touching the goal lines centrally in front of both <<Goals, goals>>. The size of the defense area is 3.6 meters times 1.8 meters for division A and 2 meters times 1 meter for division B, as shown in figures <<field-dimensions-a, 1>> and <<field-dimensions-b, 2>> respectively.
 
 
 ===== Penalty Mark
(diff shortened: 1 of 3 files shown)


tobiasheineken:
Any math on the reason why 50% was chosen?

If the field was increased by Factor 2 (which it should have for increasing to 11 robots), a mathematical way should include the goal to scale up to \sqrt(2), increasing by ~40%.

Also, as the game still has one - and only one - keeper, we should be careful when enlarging the goal too much.

g3force:
What's the reason why the field should have been increased by factor 2? I'd me interested in that math as well :)

40% would also be fine for me. That would be 1.68m which could be rounded to 1.7m instead of 1.8m.

The intention with increasing the goal when increasing the defense area is, that the ball has to be shot from a larger distance (but I think that was already clear).

Also, as there are currently quite few goals in the SSL, it is probably not too bad to increase the goal width a bit more.

tobiasheineken:
If our plan from Japan came true, we'd have doubled the whole field again in 2020, resulting in 16 (old) or 4 (new) cameras. That's what this section was about.

If so, we would have increased the playing field by *4 and the number of robots ~*2 - which looks exactly how area and distance should interact: `A = \pi*r*r`

I know why we need to increase the goal. I just wanted to encourage someone to do some math or some experiments, because I have no idea how it plays out.

+40% is also meant for defense area, not only goal itself :)

rhololkeolke:
We definitely have to increase the size to accommodate the additional robots. But we're not doubling the field, so no matter what we do, the relative sizes of the goals/defense area/field are going to be different. So basically its up to us which increase we use (40% vs 50% vs some other number). 

I think we should just accept this pull request as written (i.e. 50% increase). As we don't have any compelling data saying one number is better than the other, and 40% only makes sense in a historical context of the rules. Additionally 40% has the decimal rounding issues Nicolai mentioned.

andreaswendler:
There is a change missing for the definition of the penalty mark, which is still stated to be 1.2 meters from the field lines. While it could be changed to 1.8, I would find it more elegant to remove the number entirely and completely tie it to the defense area size.

g3force:
oh, good catch. However, with #14 the penalty mark is replaced by the the position for one-on-one anyway, when I remember correctly.

andreaswendler:
I ran long simulations against ourselves with the new proposed defense area and goal size and 11 robots and could not find any large detrimental effects. I therefore think that the 50% increase is likely fine (at least for us). It could lead to slightly more goals, but I consider that to be positive.

rolfvdhulst:
I agree with most other comments here.  I do want to make an argument for erring on the larger side of the goals (I think the proposed 50% is good).

Bigger goals tend to reward teams which have more ball possession and will typically reward teams which have an better AI when scoring is easier and increase discrepancies between teams. Whether or not this is a good thing I leave up to you, but in my personal opinion it's a good thing. This will create incentive for teams to work on their AI algorithms and more clearly reward teams for developing good software, which is definitely in line with the league's goals.

Also, it is simply more entertaining to watch games with more goals.

PR #31, opened 2020-02-16
Tune ball placement procedure

g3force:
From the rule proposal:

> The ball placement procedure has some corner cases and can be optimized in speed.
> There will be some optimizations, that have yet to be determined in detail:
> Relax the precision where possible
> Avoid ball placement at all, if the ball is already at a good position
> Be more strict with placement failures (turn it off faster)

This change implements:

* Conditions, when no ball placement is required (ball placement can not be successfully for a team automatically anymore, which would have reset the failure counter before)
* Failure counter is not reset on success, but only decremented

Some rational about what has not changed:

* Placement tolerance of 0.15m: Increasing this tolerance requires altering other margins as well, like the distance to the field lines for the placement position. As long as the tolerance is not hard to achieve, we should rather keep it as is.

### DIFF
--- a/chapters/refereecommands.adoc
+++ b/chapters/refereecommands.adoc
@@ -69,9 +69,9 @@ When the ball goes <<Ball In And Out Of Play, out of play>>, the following rules
 
 NOTE: The ball may still be moving when the placement command is issued.
 
-NOTE: If the ball is placed near to the defense area, teams are advised to keep the defense as close as possible to their target position.
-The game is continued right after the ball has been placed successfully.
-It is allowed to enter the defense area during ball placement.
+NOTE: The game commences directly after ball placement. The team receiving the ball may shoot immediately and leave the opposing team little time to arrange defensive actions if needed.
+
+NOTE: It is allowed to enter the defense area during ball placement.
 
 Ball placement is mandatory for all teams in division A.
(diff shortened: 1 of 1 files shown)


chapters/refereecommands.adoc
```diff
@@ -53,13 +64,15 @@ When the ball goes <<Ball In And Out Of Play, out of play>>, the following rules
...
+. When a teams placement failure counter reached 5, it is not allowed to place the ball for the rest of the game half. All free kicks that were a result of the ball leaving the field, are awarded to the opposing team. For all other rule violations or when both teams failed to place the ball, the ball is placed by the <<Referee, referee>>
 . If no team can place the ball, the ball is placed by the <<Referee, referee>> or the <<Assistant Referee, assistant referee>>. Both the referee as well as the assistant referee are advised to use a so-called ball handler (a long, preferably black stick-like device) to move the ball.
 
-NOTE: If the ball is already at the placement position (for example after a failed free kick by the other team), the ball placement can be successful without a robot manipulating the ball. This could lead to the counter for failed attempts resetting without a contribution by the placing team.
-
 NOTE: The ball may still be moving when the placement command is issued.
 
+NOTE: If the ball is placed near to the defense area, teams are advised to keep the defense as close as possible to their target position.
```
andre-ryll: Firstly: I totally don't get this note.
Secondly: is "defense" a technical term that is well defined? Or is it just some robots that seem to be defending something?
Thirdly: What is the "target position"?

g3force: Formulation adapted

PR #25, opened 2020-02-02
Excessive dribbling penalties

rhololkeolke:
No stoppage occurs. Increase foul counter. Goals within 2 seconds of end of offense do not count.

See https://docs.google.com/spreadsheets/d/1LfNNrE4ld0IkRGXSkKSKOnPDipdKbuH-InZJY2kATs4/edit?usp=sharing

PR #32, opened 2020-03-03
Restructure fouls and offenses sections

rhololkeolke:
- Reorganized rule violations into the following categories:
  - Offenses
  - Fouls
    - Stopping Fouls
    - No Stop Fouls
  - Ball Placement Violations
  - General Violations
- Removed advantage rule. Obsoleted by removal of stoppage to:
  - crashing
  - attacker touches robot in opponent defense area
- Added a No Goal Timer for each team
  - Every no stop foul adds 2 seconds to timer
  - Timer resets on:
    - Successful goal by opposing team
    - Half Time and overtime
- The following fouls have been changed to No Stop Fouls
  - Attacker In Defense Area
  - Excessive Dribbling
  - Ball Speed
  - Defender Too Close To Ball
  - Attacker Touches Robot In Opponent Defense Area
- Lack of progress timer changes
  - Decrease timeout for some situations in Div A

Feedback is specifically requested on:
- Any changes/issues with the categories.
- Should any of violations be in a different category?
- No goal timer
  - Some initial discussion on #25, #23, 
  - Is the time added for violations appropriate?
  - Should there be a timer cap?
- No Stop Foul grace period
  - Is the amount of time before triggering another no stop foul appropriate?

Other open foul pull requests will be closed in favor of this one as they will depend on the restructuring changes in this pull request.

### DIFF
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -47,7 +47,7 @@ Chapter <<Scoring Goals>>:
 | Invalid Goal | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
 |===
 
-Chapter <<Rule Violations>>, section <<Stopping Offenses>>:
+Chapter <<Offenses>>, section <<Stopping Offenses>>:
 |===
 | Event | Applicability | Command | AutoRef
 
@@ -59,7 +59,7 @@ Chapter <<Rule Violations>>, section <<Stopping Offenses>>:
 |===
 
 
-Chapter <<Rule Violations>>, section <<Stopping Fouls>>:
+Chapter <<Offenses>>, section <<Stopping Fouls>>:
 |===
 | Event | Applicability | Command | AutoRef
 
@@ -74,7 +74,7 @@ Chapter <<Rule Violations>>, section <<Stopping Fouls>>:
 |===
 
 
-Chapter <<Rule Violations>>, section <<No Stop Fouls>>:
+Chapter <<Offenses>>, section <<No Stop Fouls>>:
 
 |===
 | Event | Applicability | Command | AutoRef
@@ -89,7 +89,7 @@ Chapter <<Rule Violations>>, section <<No Stop Fouls>>:
 | <<Crashing>> draw | always | no command | icon:check[role="green"]
 |===
 
-Chapter <<Rule Violations>>, section <<General Violations>>:
+Chapter <<Offenses>>, section <<General Violations>>:
 
 |===
 | Event | Applicability | Command | AutoRef
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -70,23 +70,24 @@ Chapter <<Rule Violations>>, section <<Stopping Offenses>>:
 Chapter <<Reul Violations>>, section <<Stopping Fouls>>:
 |===
 | Event | Applicability | Command | AutoRef
-| Multiple <<Fouls>> | <<Ball In And Out Of Play, ball out of play>> |
-  <<Yellow Card>> | icon:times[role="red"] (handled by the game
-  controller)
-| <<Robot Too Close To Opponent Defense Area>> | <<Ball In And Out Of Play, ball out of play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
+
+| <<Robot Too Close To Opponent Defense Area>> | during <<Stop>> and <<Free Kick>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
 | <<Pushing>> | always | <<Stop>>, then <<Free Kick>> | icon:times[role="red"]
 | <<Ball Holding>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
 | <<Tipping Over Or Dropping Parts>> | always | <<Stop>>, then <<Free Kick>> | icon:times[role="red"]
 | <<Robot Stop Speed>> | during <<Stop>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
 | <<Multiple Defenders>> partially | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>>, <<Yellow Card>> | icon:check[role="green"]
 | <<Multiple Defenders>> entirely | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Penalty Kick>> | icon:check[role="green"]
-| <<Boundary Crossing>> | <<Ball In And Out Of Play, ball in play>> |<<Stop>> then  <<Free Kick>> | 
+| <<Boundary Crossing>> | <<Ball In And Out Of Play, ball in play>> |<<Stop>> then  <<Free Kick>> |
 |===
 
 
-Chapter <<Rule Violations>> section <<No Stop Fouls>>:
+Chapter <<Rule Violations>>, section <<No Stop Fouls>>:
 
 |===
+| Event | Applicability | Command | AutoRef
+
+| Multiple <<Fouls>> | <<Ball In And Out Of Play, ball out of play>> | <<Yellow Card>> | icon:times[role="red"] (handled by the game controller)
 | <<Attacker In Defense Area>> | <<Ball In And Out Of Play, ball in play>> | <<No Stop Fouls, No goal timer>> increase | icon:check[role="green"]
 | <<Attacker Touches Robot In Opponent Defense Area>> skipped | <<Ball In And Out Of Play, ball in play>> | <<No Stop Fouls, No goal timer>> increase | icon:check[role="green"]
 | <<Excessive Dribbling>> | <<Ball In And Out Of Play, ball in play>> | <<No Stop Fouls, No goal timer>> increase| icon:check[role="green"]
@@ -96,7 +97,7 @@ Chapter <<Rule Violations>> section <<No Stop Fouls>>:
 | <<Crashing>> draw | always | no command | icon:check[role="green"]
 |===
 
-Chapter <<Offenses>>, section <<Unsporting Behavior>>:
+Chapter <<Rule Violations>>, section <<General Violations>>:
 
 |===
 | Event | Applicability | Command | AutoRef
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -58,7 +58,7 @@ NOTE: Note that the information shown in this table here is incomplete. Please r
 === Events For Stopping Fouls
 All <<Stopping Fouls>> result in:
 
-* Incrementing the foul counter.
+* Incrementing the foul counter (except for <<Multiple Defenders>>).
 
 |===
 | Event | Applicability | Action | Handled By
@@ -67,10 +67,10 @@ All <<Stopping Fouls>> result in:
 | <<Pushing>> | always | <<Stop>> -> <<Free Kick>> | referee -> game controller
 | <<Ball Holding>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | referee -> game controller
 | <<Tipping Over Or Dropping Parts>> | always | <<Stop>> -> <<Free Kick>> | referee -> game controller
-| <<Multiple Defenders>> partially | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Yellow Card>> -> <<Free Kick>> | auto referee
-| <<Multiple Defenders>> entirely | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Penalty Kick>> | auto referee
+| <<Multiple Defenders>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Penalty Kick>> | auto referee
 | <<Boundary Crossing>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
 | <<Keeper Held Ball>> in <<Defense Area, defense area>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
+| <<Excessive Dribbling>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
 |===
 
 
@@ -78,13 +78,12 @@ All <<Stopping Fouls>> result in:
 All <<Non Stopping Fouls>> result in:
 
 * Incrementing the foul counter.
-* Increasing the no goal timer.
+* Repeated every 2 seconds, if still committed without interruption.
 
 |===
 | Event | Applicability | Action | Handled By
 
-| <<Attacker In Defense Area>> | <<Ball In And Out Of Play, ball in play>> | - | auto referee
-| <<Excessive Dribbling>> | <<Ball In And Out Of Play, ball in play>> | - | auto referee
+| <<Attacker Touched Ball In Opponent Defense Area>> | <<Ball In And Out Of Play, ball in play>> | - | auto referee
 | <<Ball Speed>> | <<Ball In And Out Of Play, ball in play>> | - | auto referee
 | <<Attacker Touches Robot In Opponent Defense Area>> | <<Ball In And Out Of Play, ball in play>> | - | auto referee
 | <<Crashing>> | always | - | auto referee
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -30,7 +30,7 @@ NOTE: Note that the information shown in this table here is incomplete. Please r
 | <<Throw-In>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
 | <<Goal Kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
 | <<Corner Kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
-| <<Aimless Kick [small]#(_division B only_)#, Aimless Kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
+| <<aimless-kick>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
 |===
 
 
@@ -117,5 +117,5 @@ This is a complete list of differences between <<Divisions, division>> A and <<D
 * Division A plays on a <<Dimensions, larger field>> with <<Goals, larger goals>> than division B. As a result, the <<Shoot-Out, shoot-out>> is taken from a greater distance as well.
 * Division A plays with <<Number Of Robots, more robots>> than division B.
 * The automatic <<Ball Placement, ball placement>> procedure is mandatory for division A and optional for division B.
-* The <<Aimless Kick [small]#(_Division B only_)#, aimless kick>> rule only applies to division B.
+* The <<aimless-kick, aimless kick>> rule only applies to division B.
 * Division A has shorter timeouts in some situations
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -47,11 +47,11 @@ NOTE: Note that the information shown in this table here is incomplete. Please r
 |===
 | Event | Applicability | Command | Handled By
 
-| Multiple <<Yellow Card, Yellow Cards>> | always | <<Read Card>> | game controller
 | <<No Progress In Game>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Force Start>> | auto referee
 | <<Double Touch>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>> -> <<Free Kick>> | auto referee
-| Multiple <<Fouls>> | always | <<Yellow Card>> | game controller
 | <<Unsporting Behavior>> | always | <<Stop>> -> <<Yellow Card>>, <<Red Card>>, <<Penalty Kick>>, <<Forced Forfeit>> or <<Disqualification>> | referee -> game controller
+| Multiple <<Fouls>> | always | <<Yellow Card>> | game controller
+| Multiple <<Yellow Card, Yellow Cards>> | always | <<Red Card>> | game controller
 |===
 
 
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -1,13 +1,5 @@
 [appendix]
 == Terminology
-=== Ball In And Out Of Play
-When the match is <<Stopping The Game, stopped>>, the ball is considered *out of play* until it has been brought into play.
-
-When the match is <<Resuming The Game, resumed>>, the ball is considered *in play* until the next stoppage occurs. The match is resumed when <<Force Start, force start>> has been issued or the ball moved at least 0.05 meters following a <<Kick-Off, kick-off>>, <<Free Kick, free kick>> or <<Penalty Kick, penalty kick>>.
-
-NOTE: see <<Double Touch, double touch>> for the rationale of the 0.05 meter distance
-
-
 === Ball Manipulation
 Shooting and <<Dribbling Device, dribbling>> is considered as manipulating the ball, the ball accidentally bouncing off the hull is not.
 
--- a/chapters/appendix.adoc
+++ b/chapters/appendix.adoc
@@ -74,8 +74,8 @@ All <<Stopping Fouls>> result in:
 |===
 
 
-=== Events For No Stop Fouls
-All <<No Stop Fouls>> result in:
+=== Events For Non Stopping Fouls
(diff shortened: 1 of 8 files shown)


chapters/appendix.adoc
```diff
@@ -55,38 +55,45 @@ Chapter <<Scoring Goals>>:
...
 | <<Double Touch>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
-| <<Attacker In Defense Area>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
-| <<Attacker Touches Robot In Opponent Defense Area>> skipped | <<Ball In And Out Of Play, ball in play>> | no command | icon:check[role="green"] (<<Advantage Rule>>)
-| <<Excessive Dribbling>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
-| <<Ball Speed>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
 |===
 
 
-Chapter <<Offenses>>, section <<Fouls>>:
+Chapter <<Reul Violations>>, section <<Stopping Fouls>>:
```
tobiasheineken:
```suggestion
Chapter <<Rule Violations>>, section <<Stopping Fouls>>:
```

chapters/appendix.adoc
```diff
@@ -55,38 +55,45 @@ Chapter <<Scoring Goals>>:
 | Invalid Goal | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
 |===
 
-Chapter <<Offenses>>, section <<Minor Offenses>>:
+Chapter <<Rule Violations>>, section <<Stopping Offenses>>:
```
tobiasheineken:
```suggestion
Chapter <<Rule Violations>>, section <<Non Stopping Offenses>>:
```

chapters/appendix.adoc
```diff
@@ -55,38 +55,45 @@ Chapter <<Scoring Goals>>:
...
 
 
-Chapter <<Offenses>>, section <<Fouls>>:
+Chapter <<Reul Violations>>, section <<Stopping Fouls>>:
 |===
 | Event | Applicability | Command | AutoRef
-
-| Multiple <<Fouls>> | <<Ball In And Out Of Play, ball out of play>> | <<Yellow Card>> | icon:times[role="red"] (handled by the game controller)
-| <<Attacker Touches Robot In Opponent Defense Area>> | <<Ball In And Out Of Play, ball in play>> | <<Stop>>, then <<Free Kick>> | icon:check[role="green"]
+| Multiple <<Fouls>> | <<Ball In And Out Of Play, ball out of play>> |
```
tobiasheineken: Afaik, there won't be a stop for multiple fouls as yellow cards no longer need to interrupt the game.

chapters/appendix.adoc
tobiasheineken: I guess I misread this section. I do now get the difference between Stopping Offenses and Stopping Fouls.

I'm not sure about Lack of progress here: Why is it listed here?

chapters/offenses.adoc
```diff
@@ -1,24 +1,95 @@
...
 
-NOTE: If both teams are clearly unable to bring the ball <<Ball In And Out Of Play, into play>> without violating the rules, the referee may issue a <<Force Start, force start>> command instead of an <<Free Kick, free kick>> for the other team.
+The ball must have moved at least 0.05 meters to be considered as <<Ball In And Out Of Play, in play>>.
+
+NOTE: It is understood that the ball may be bumped by the robot multiple times over a short distance while the kick is being taken. This is why a distance of 0.05 meters is used to decide whether a robot violates this rule or not. Remaining in contact with the ball for more than 0.05 meters also counts as double touch, even though technically the robot only touched the ball once.
 
 ==== Double Touch
 When the ball is brought <<Ball In And Out Of Play, into play>> following a <<Kick-Off, kick-off>>, <<Free Kick, free kick>> or <<Penalty Kick, penalty kick>>, the kicker is not allowed to touch the ball until it has been touched by another robot or the game has been stopped.
```
tobiasheineken: Repetition?

chapters/offenses.adoc
```diff
@@ -27,126 +98,155 @@ The ball must have moved at least 0.05 meters to be considered as <<Ball In And
...
+NOTE: When the ball is <<Ball In And Out Of Play, out of play>>, the rule <<Robot Too Close To Opponent Defense Area>> applies instead.
 
-If a robot other than the keeper touches the ball while being entirely inside its own defense area, the game is stopped and a <<Penalty Kick, penalty kick>> is awarded to the other team. The foul counter is not increased.
 
-==== Boundary Crossing
-A robot must not kick the ball over the field boundary such that the ball leaves the field.
+===== Crashing
+At the moment of collision of two robots of different teams, the difference of the speed vectors of both robots is taken and projected onto the line that is defined by the position of both robots. If the length of this projection is greater than 1.5 meters per second, the faster robot committed a foul. If the absolute robot speed difference is less than 0.3 meters per second, both conduct a foul but the game will not be stopped.
```
tobiasheineken:
```suggestion
At the moment of collision of two robots of different teams, the difference of the speed vectors of both robots is taken and projected onto the line that is defined by the position of both robots. If the length of this projection is greater than 1.5 meters per second, the faster robot committed a foul. If the absolute robot speed difference is less than 0.3 meters per second, both conduct a foul.
```

chapters/offenses.adoc
```diff
@@ -27,126 +98,155 @@ The ball must have moved at least 0.05 meters to be considered as <<Ball In And
...
+NOTE: Dribblers can still be used to dribble large distances with the ball as long as the robot periodically loses possession, such as kicking the ball ahead of it as human soccer players often do.
 
-==== Defender Too Close To Ball
-A robot's distance to the ball must be at least 0.5 meters during an opponent <<Kick-Off, kick-off>> or <<Free Kick, free kick>>.
+===== Ball Speed
+A robot must not accelerate the ball faster than 6.5 meters per second in 3D space.
+
+===== Defender Too Close To Ball
+A robot's distance to the ball must be at least 0.5 meters during an opponent <<Kick-Off, kick-off>>, <<Free Kick, free kick>>.
 The game is resumed with the same command that was issued before the foul.
```
tobiasheineken: Obsolete, if the game state does not change?

Also, this has more implications.
It's easy, especially with mediocre vision, to misidentify the ball to be "in play".
The previous rules already punished teams for misidentifying with a foul - so that's not the point.

The problem is that afaik in play / out of play has to implemented by the teams and therefore the "Stop" command to push each AI towards a stable "The ball is out of play, keep your distance" is necessary.

Is there any other hint to the AI they have to keep their distance right now?

chapters/appendix.adoc
g3force: correct, I've moved it to the other section.

chapters/offenses.adoc
g3force: removed

chapters/offenses.adoc
g3force: I've remove the line.

Teams have to do the following:
* If command is stop or prepare_kickoff -> keep distance
* If command is normal_start or direct_free_kick -> wait until ball moved or timeout counter got below 0 (contained in referee msg)

Now that we do not stop for this rule, what can happen?
* Opponent is just a bit too close to the ball -> team is punished, attacking team should have no significant disadvantage
* Opponent is approaching the ball -> team is punished, attacking team can not perform free kick -> should be repeated

The second case often happens due to vision issues, which the autoRefs could also suffer from. So the human ref should have an eye on these situations anyway.
I've added another sentence that allows the human ref to repeat the situation. I'd like to avoid defining this more concrete, to keep it simple in the rules and in the autoRefs.

chapters/appendix.adoc
g3force: I've refactored the Stopping Offenses away

tobiasheineken:
How about resetting the no-goal-timer on any free kick / Stop aswell?
Otherwise we should at least write an exemption for penalties.

g3force:
Yeah, agree. It is easier and more reasonable to reset it when the game is stopped. The idea is to avoid scoring goals right after committing fouls.

g3force:
Currently, the timer is defined as: Add 2 seconds to the timer with each non-stopping foul.
I would rather choose something like: Set the timer to 5 seconds on each non-stopping foul.

g3force:
I've created a new sheet in the original spreadsheet with an up-to-date version of all the events for a better overview: https://docs.google.com/spreadsheets/d/1LfNNrE4ld0IkRGXSkKSKOnPDipdKbuH-InZJY2kATs4/edit#gid=1428240242

g3force:
Current HTML version and HTML diff of this PR compared to master for easier review:

[rules.zip](https://github.com/RoboCup-SSL/ssl-rules/files/4302294/rules.zip)

andre-ryll:
I would try to avoid the "No Goal Timer" if possible. A timer that is running in the background makes thing quite complicated and also can lead to large confusion. Just think about a crash somewhere on the field while a goal is shot. The goal wouldn't count.

For "Ball Speed", "Attacker In Defense Area", and "Attacker Touches Robot In Opponent Defense Area" I would simply validate these three conditions if a (possible) goal was shot. If before the shot any condition was true the goal will not count.

"Defender Too Close To Ball" won't lead to a goal soon anyway for the defending team.

Most difficult is "Excessive Dribbling". I recommend to either give a yellow card immediately and keep the game running or make it a foul where the game is stopped. It didn't happen too often in the games anyway.

This way we can remove the "No Goal Timer" and simplify things.

andre-ryll:
Thinking a bit more about the rules I would also recommend to remove the rules "Attacker Touches Robot In Opponent Defense Area" and "Multiple Defender Partially" to simplify the game.

Currently, we have  a lot of discussions during a game due to actions very close at the defense area. Did a defender enter the defense area? Was it pushed? Did it have ball contact? Did the ball in turn have attacker contact?
Depending on the answer to all these questions different actions follow. It is very complicated to see by a human ref all at the same time and the autoref may have issues in such situations as the ball is not visible.

I assume the defense area was introduced to prevent a team from placing all robots in front of the goal to effectively block it. Given this as the main objective of the defense area all the details of actions directly on the line are not relevant.

What may happen if the two above listed rules are dopped:
- Defenders may fall back slightly during a close-contact situation with the ball without being afraid of penalties
- It is much easier to judge fouls as only "Attacker in Defense Area" and "Multiple Defender Fully" remains. Pushing is easier to see for the human ref as the robots may move back up to approx. 10cm before a game is stopped due to multiple defenders.
- A team that permanently puts defenders inside the penalty area (partially) would get "unsportive behaviour" fouls.
- It is allowed to partially enter the defense area in case of imprecise control of the robots, this will not be punished.

Overall this relaxes the field line meaning of the defense area while still fullfilling the main purpose not to be able to block the goal. It makes ref life easier, for the audience the game is easier to follow, we have less rules to check and less stoppages.

g3force:
I would prefer to move the excessive dribbling rule back to the stopping fouls. It did not happen often and this way there is no special case.

So, to sum up the non-stopping fouls that would currently effect the no-goal timer:

Attacker [touched ball] in [opponent] defense area: No goal, if ball kicked from within defense area
Ball Speed: No goal, if goal shot was too fast
Defender Too Close to Ball: not relevant for goal shots
Crashing: not relevant for goal shots, or punishable by unsporting behavior
Robot Stop Speed: not relevant for goal shots
Ball Placement Interference: not relevant for goal shots
Excessive Dribbling: Convert to stopping foul

That sounds good to me :)

andreaswendler:
Regarding the second comment of @andre-ryll 
I agree with you regarding the "Attacker Touches Robot In Opponent Defense Area" rule. While it was introduced to prevent pushing, it mostly occurred in other (completely harmless) circumstances. Removing this rule again should simplify things.
I can only partially agree with you regarding the removal of the "Multiple Defender Partially" rule. It is of course necessary to remove it for the removal of the attacker touch robot ... rule. Removing it is probably fine, but I would penalize permanent (partial) stay of the defenders in the defense area with a precise rule, not just unsporting behavior (seeing that I do not now how that is really unsporting).
For the clarity of the game (both for spectators and teams), I find it important for the robots to stay out of the marked defense area at least most of the time.
A rule to prevent this could for example forbid any continuous stay inside the defense area for more than 2 seconds for any robot.

g3force:
It should definitely be stated somewhere, that robots should not constantly stay (partially) inside the defense area. Then, the human ref also has a reason for the unsporting behavior.

An explicit rule would be an option, though I'm not sure if a fixed time limit is the right metric. 2 seconds seems definitely too short to me. I think, a human ref can better judge, if a team moves inside the defense area consistently on purpose.

andre-ryll:
Honestly, I think there is no proper metric we would all agree on. A rule infraction on that topic is very much "intention-based". And that is hard to capture with numbers.

Touching the defense area while moving because robot control is not the best? That's okay.
A team new to Div A staying permanently in the defense area with one robot with just a centimeter? That's okay. They are not doing it intentionally.
A robot moving through the defense area when the ball is not even close? That's okay.
As we see there are so many case-by-case decisions that it's better to leave it to the human ref to judge it unsporting. That also gives most freedom of the penalty. Like giving a warning first or ask the team to fix the issue in a timeout. If the team is in general not the best be a bit more relaxed.

I would just state something like "Robots other than the keeper must maintain best-effort to fully stay outside the defense area. Infraction of this rule can be rated as unsporting behavior.".

rhololkeolke:
> Currently, the timer is defined as: Add 2 seconds to the timer with each non-stopping foul.
> I would rather choose something like: Set the timer to 5 seconds on each non-stopping foul.

I agree that reset to 5 seconds is better than unlimited add 2 seconds to the timer.

> How about resetting the no-goal-timer on any free kick / Stop aswell?
> Otherwise we should at least write an exemption for penalties.

I agree. Free kicks should clear the timer as well.



> Honestly, I think there is no proper metric we would all agree on. A rule infraction on that topic is very much "intention-based". And that is hard to capture with numbers.
> 
> Touching the defense area while moving because robot control is not the best? That's okay.
> A team new to Div A staying permanently in the defense area with one robot with just a centimeter? That's okay. They are not doing it intentionally.
> A robot moving through the defense area when the ball is not even close? That's okay.
> As we see there are so many case-by-case decisions that it's better to leave it to the human ref to judge it unsporting. That also gives most freedom of the penalty. Like giving a warning first or ask the team to fix the issue in a timeout. If the team is in general not the best be a bit more relaxed.
> 
> I would just state something like "Robots other than the keeper must maintain best-effort to fully stay outside the defense area. Infraction of this rule can be rated as unsporting behavior.".

If we want to keep rigidly defined rules, perhaps we can relax them a bit by only applying these offenses when the ball is on the defending team's half of the field? It would allow for scenarios like you suggested (e.g. moving through defense area) when there is clearly no immediate defensive benefit.

The ref always has the discretion to call things like unsporting behavior. So even if we have explicit rules, they can still call unsporting behavior if the spirit of the law is violated. Regardless, I think we should add something lke "Robots other..." so that everyone reading and enforcing the rules is on the same page as to the intent of the rules.

tobiasheineken:
@andre-ryll 

Honestly, I disagree with your comments a lot. The "No Goal Timer" was introduced because 

> We should use the grace period here as well. Making a fast pass and than scoring a goal should also not be allowed.

(See #23).
While it is confusing, and I'm wishing for a better solution (Stop -> Freekick is superior in this regard: If you ignore the fact that you don't understand **why** the freekick happend, you still know **what** will happen now, as it's a usual way to continue in many sports), only considering these rules for the goal-shot itself doesn't seem right to me. And while I myself complained "a state of the game where one team doesn't want to shoot a goal could be a problem", I agree with Nicolai "it could also be interesting to keep in possession until you're able to score again". The mayor drawback here is implementation work.

And relying on "unsporting behavior" to solve "intention based infringements" is lazy rule design (Re: Crashing, staying inside the defense area, ...) . On the one hand it's by design impossible to resolve automatically (which is bad for a robotics league), on the other hand it's also guesswork for the teams. Are you allowed to cut in the defender's path when he tries to go for the ball? With _Attacker Touches Robot In Opponent Defense Area_ you're not (at least close to the defense). With _Crashing_ you know that you could invalidate your shot by doing so, if you end up faster. A risk worth taking - or maybe not - but it's up to the teams. If we leave it open, it's impossible to do this risk analysis. Maybe one referee will issue a yellow card, one referee will force you to take a timeout and remove the feature, one referee will say it's ok. How shall we play with these kind of rules? Also: While crashing while being faster needs some punishment imo, being slower and blocking the way for a robot moving towards its defense area sounds fair to me.

Also, Regarding @andreaswendler: While _Attacker Touches Robot In Opponent Defense Area_ was invented to solve a pushing issue, it forcing the attacker to keep some distance for any defender  is a good thing. Robots are simply not capable like humans in handball to "fight" for their position on the line.

tjones320 (commented):
I really like this new organizational structure. I will examine it more later, but I think it's pretty solid. I wonder how flexible we are about changing some of the things off of auto ref. I remember last year there were a few calls that it was better not to let the computer decide. I think double touch may have been a good example? Maybe have a note that reminds people that auto-ref calls can be changed to manual after an agreement of both teams. Maybe I'm mistaken about the current ideas about auto-ref. Otherwise, this is really good work! Great job!

chapters/offenses.adoc
```diff
@@ -1,152 +1,169 @@
...
+* not obeying the orders of the <<Referee, referee>>
 
 === Fouls
-Fouls are violations of the rules that result in a <<Free Kick, free kick>> for the opposing team. The free kick will be shot from the position where the ball was located when the offense began happening (see <<Free Kick, the free kick rules>> for the exact ball position rules). If the foul happened while the ball is <<Ball In And Out Of Play, out of play>>, no free kick is given.
+The number of fouls per team is tracked by a counter. Each foul will
+increase the counter by one. Every third increase to the foul counter
+causes a <<Yellow Card, yellow card>> to be awarded.
 
-Every third foul of the same team results in a <<Yellow Card, yellow card>>.
+Every violation in this section and its subsections increase the foul
```
andreaswendler: This is not technically true since multiple defender does not increase the foul counter. It is however very explicitly written there that it does not

chapters/offenses.adoc
```diff
@@ -1,152 +1,169 @@
...
+===== Ball Speed
+A robot must not accelerate the ball faster than 6.5 meters per second in 3D space.
 
-==== Damaging Other Robots
-It is not allowed to damage or modify robots of other teams.
+The <<Referee, human referee>> may decide to repeat the <<Kick-Off, kick-off>> or <<Free Kick, free kick>> on significant disturbances.
 
-==== Damaging The Field Or The Ball
-It is not allowed to damage or modify the field or the ball.
+NOTE: During <<Stop, stop>>, there is no automatic sanction for being too close to the ball. The referee may still punish a team for <<Unsporting Behavior,unsporting behavior>> by issuing a <<Yellow Card, yellow card>> if it does not respect the required distance. See <<Stop, stop>> for further explanation.
```
andreaswendler: This note seems to be really out of place here. It does not relate to the ball speed foul at all. If there is no accompanying foul, it might still be better to put it in the Fouls while ball out of play section

chapters/offenses.adoc
```diff
@@ -1,152 +1,169 @@
...
 
-==== Showing Lack Of Respect
-A team member must show appropriate respect to everyone involved in the game. Infringements of this rule include but are not limited to:
+===== Attacker Touches Robot In Opponent Defense Area
+When the ball <<Ball In And Out Of Play, in play>>, a robot must not touch any opponent robot inside the opponent <<Defense Area, defense area>>.
```
andreaswendler:
```suggestion
When the ball is <<Ball In And Out Of Play, in play>>, a robot must not touch any opponent robot inside the opponent <<Defense Area, defense area>>.
```

chapters/playingenvironment.adoc
```diff
@@ -81,7 +81,7 @@ Besides the shared vision equipment, teams are not allowed to mount their own ca
 A game is controlled by the community-maintained ssl-game-controller (https://github.com/RoboCup-SSL/ssl-game-controller).
 It is operated by the <<Game Controller Operator, game controller operator>>. The software translates decisions of the <<Referee, referee>> and the <<Automatic Referee, automatic referee>> into Ethernet communication signals that are broadcast to the network. It maintains the state of the game, tracks all events and acts as a proxy between all participating parties in the game.
 
-The game-controller has a network interface for the playing teams. They can automatically <<Choosing Keeper Id, change their keeper id>>, they can signal a robot substitution intent for the next opportunity and they can reply to requests of the <<Advantage Rule, advantage rule>>.
+The game-controller has a network interface for the playing teams. They can automatically <<Choosing Keeper Id, change their keeper id>>, they can signal a robot substitution intent for the next opportunity.
```
andreaswendler:
```suggestion
The game-controller has a network interface for the playing teams. They can automatically <<Choosing Keeper Id, change their keeper id>> and they can signal a robot substitution intent for the next opportunity.
```

chapters/offenses.adoc
```diff
@@ -1,152 +1,169 @@
...
-Fouls are violations of the rules that result in a <<Free Kick, free kick>> for the opposing team. The free kick will be shot from the position where the ball was located when the offense began happening (see <<Free Kick, the free kick rules>> for the exact ball position rules). If the foul happened while the ball is <<Ball In And Out Of Play, out of play>>, no free kick is given.
-
-Every third foul of the same team results in a <<Yellow Card, yellow card>>.
+The number of fouls per team is tracked by a counter. Each foul will
+increase the counter by one. Every third increase to the foul counter
+causes a <<Yellow Card, yellow card>> to be awarded.
 
-In case of severe fouls, the referee can also issue a <<Yellow Card, yellow card>> or a <<Red Card, red card>>.
+Every violation in this section and its subsections increase the foul
+counter.
```
```suggestion
Violations in this section and its subsections increase the foul counter if not stated otherwise.
```

chapters/offenses.adoc
g3force: I've added a suggestion to formulate it more correctly.

chapters/offenses.adoc
g3force: good catch. This and the sentence before belongs to the Defender Too Close To Ball section. I've moved it.

g3force:
@tjones320 I added a sentence about disabling game events.

@rhololkeolke 

> If we want to keep rigidly defined rules, perhaps we can relax them a bit by only applying these offenses when the ball is on the defending team's half of the field? It would allow for scenarios like you suggested (e.g. moving through defense area) when there is clearly no immediate defensive benefit.

Right now, the defender has to touch the ball already, so the ball is obviously nearby. Before 2014, it was actually sufficient that the ball was inside the defense area, which was even harder and led to more penalty kicks.

@tobiasheineken I can understand (at least some of) your concerns. Though there is always a trait-off  between rule complexity and covered corner cases.

I have the impression that you want to implement your strategy as close to the rule boundaries as possible to get the most out of the possibilities. And for that you need precise rules, obviously. But you could also just assume a safe tolerance. For example, for the removal of the partially inside defense area rule, why not assuming that you are still not allowed to ever enter the defense area? That is definitely a best effort and you can still compensate imprecise movement of your robots.

g3force:
I've implemented the suggested changes for the no-goal-timer replacement.

Latest version of the rules and the diff:

[rules.zip](https://github.com/RoboCup-SSL/ssl-rules/files/4331369/rules.zip)

tobiasheineken:
The version with the no-goal-timer was superior. But even if we want to live without it, we need to address some issues:

A) _Attacker Touches Robot in Defense Area_: We need some variation of that rule. Without it, the attacker can simply place itself on the border of the penalty area and accept a chip-shot and there's no defender apart from the keeper who is able to challenge this.
B) _Fast passes_: Nicolai stated himself that we don't want to allow a very fast pass into a goalshot, even if the goalshot itself is slow.
C) _Related crashes_: There are situations where a crash is directly related to a goal-shot, in particular blocking the defender to reach the ball. These situations are **not unsporting per se**. It's just that I miscalculated my deceleration and didn't come to stop in time. We cannot accept this to happen. We cannot rely on the referee to handle it properly because we all know they won't. We need some kind of clarification.

And, unrelated: I never tried to avoid buffer-zones. We use them all over the place to compensate for imprecise movement. But it's not the rule's job to define them for us. You're not allowed to touch the ball partially inside the defense area is a perfectly safe rule. We keep some additional distance, just in case. Imprecise motion is on us. We can work on it, we can improve it and we can reduce the additional distance if me made an improvement. "You **are** allowed to touch the ball inside your defense area if at least one part of your robot is outside" is also a perfectly useful rule. Again, imprecise motion is on us. But this is - meh. You're actively incentivize to have a lot of imprecise motion here. Beeing closer to the goal is an obvious advantage and instead of rewarding precise movement (which ER-Force don't have, don't get me wrong here) by enabling a team with more precise motion to reduce their additional distance, we reward you to claim that your robot's motion is not that good. Like in the other PR, this is not about "intentional making it worse" but "stop improving" because it does no longer matter.

g3force:
I fine with the rules as it is written now, regarding the points A-C. Does anyone else have similar concerns to those @tobiasheineken raised?

PR #40, opened 2021-02-10
Clarify how distances are measured between lines

### LINKED ISSUE #33: Consistent use of line thickness
It was brought up on Discord, that there is some confusion in regards of the line thickness:

* How are distances between lines measured?
* How is the line thickness taken into account when checking if object are inside an area?
* When did the ball cross a line?

Current state:

* ssl-rules
  * 2.1.3: > Lines belong to the areas of which they are boundaries.
* ssl-vision
  * Distances between lines are measured from the line center
  * Line thickness is configurable and sent out in the Geometry message
  * Line thickness is only taken into account during calibration
* ssl-game-controller
  * line thickness is not used and not needed
* TIGERs autoRef
  * The line thickness is not used
* ER-Force autoRef
  * The line thickness is most likely not used

Some relevant quotes from the rules:

* "The ball is inside the field."
* "allowed to be inside the whole center circle"
* "the ball is inside a team’s defense area"
* "while being partially or fully inside the opponent defense area"
* "The robot is at least partially inside the field margin."

The line is usually 1cm thick. The precision of ssl-vision is also about 1cm. So in practice, it should not make a significant difference to take the thickness into account or not.

g3force:
Suggested change to clarify how distances on the field should be measured.

This is, how ssl-vision is doing it and how probably most teams interpreted it.

closes #33

### DIFF
--- a/chapters/playingenvironment.adoc
+++ b/chapters/playingenvironment.adoc
@@ -9,9 +9,9 @@ The field of play must be rectangular and of the following size:
 
 The exact field dimensions and the field markings at the venue may vary by up to ±10% in each linear dimension.
 
-The two figures below show the dimensions of the field, the goals and special field areas, measured in millimeters.  <<field-dimensions-a, Figure 1>> shows the dimensions for division A and  <<field-dimensions-b, figure 2>> for division B.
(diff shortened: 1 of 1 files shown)


lhcavalcanti (approved):
👍

PR #47, opened 2022-01-19
Use backend-html5 instead of basebackend-html

kkimurak:
The backend of `asciidoctor-pdf` is [also html](https://github.com/asciidoctor/asciidoctor-pdf/blob/v1.6.x/lib/asciidoctor/pdf/converter.rb#L121) so `basebackend-html` directive will be unintentionally displayed even in the `asciidoctor-pdf` artifact.
Using backend-html5 directive, what shouldn't be displayed will not be displayed as intended. 

snapshot-20220120utc+9_unintentionally-displayed-content

### DIFF
--- a/sslrules.adoc
+++ b/sslrules.adoc
@@ -6,11 +6,11 @@
 :toclevels: 4
 
 // add icons from fontawesome in a up-to-date version
-ifdef::basebackend-html[]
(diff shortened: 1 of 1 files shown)


g3force:
Looks good, thx :+1:

PR #49, opened 2022-03-31
Bugfix: Goals should be invalid, if bot committed foul

g3force:
Right now, a bot could touch the ball inside the penalty area (committing non-stopping foul) and afterwards kick the ball into the goal. It would be valid, because the foul was before the robot last touched the ball. A 2-second timer will avoid this. 2 seconds is also the grace period for repeating a foul that is still being committed.

Respective change in the GC: https://github.com/RoboCup-SSL/ssl-game-controller/pull/23/files


g3force:
Decision in today's OC/TC meeting: This is a bug in the rules and we want to have the fix for this year. It should not require teams to change their software.

PR #51, opened 2022-05-06
fix broken internal cross reference

kkimurak:
Due to a syntax error, an internal cross reference is unintentionally rendered as is.
cross-reference for Robot Handler rendered as is

This PR fixes it. 

This should have been done during the review period.. I'm sorry I'm late to notice.


g3force:
oops, thx for the fix :)

PR #52, opened 2022-07-10
Bugfix: aimless kicks must cross the halfline, not the midline

chrismit3s:
As discussed on the discord:

> I have a question about the aimless kick rule (6.2.3) in division B. Under Usage it states: "A kick is aimless when after the ball touched a robot, it subsequently crossed the midline and then its opponent’s goal line outside the goal without touching another robot." The midline is defined as the line parallel to the touch lines (the long borders of the field).
>
> Say a bot stood close to the defense area of the opponent, but slightly to the left of the midline. If that bot now shot at the goal, but missed it and the ball left the field to the right of the goal via the goalline, that would qualify as an aimless kick.
>
> Is this correct? Personally, I think the halfway line would be better suited for this definition.

### DIFF
--- a/chapters/ballleavesthefield.adoc
+++ b/chapters/ballleavesthefield.adoc
@@ -44,6 +44,6 @@ The ball has to be placed at the position from where the ball was kicked (see th
 After the ball has been placed, a <<Free Kick, free kick>> is awarded to the opponent of the team that last touched the ball before it left the field.
 
 .Usage
-A kick is aimless when after the ball touched a robot, it subsequently crossed the midline and then its opponent's goal line outside the goal without touching another robot.
+A kick is aimless when after the ball touched a robot, it subsequently crossed the halfline and then its opponent's goal line outside the goal without touching another robot.
 
 A kick-off kick cannot be aimless, as the ball is located at the midline and does therefore not cross it.


g3force (approved):
Looks good to me.

In the long term, we could also define the mid line and half way line explicitly and reference the definition.

I'll wait a bit to see if there are any concerns with this change, but I doubt there will.

chapters/ballleavesthefield.adoc
```diff
@@ -44,6 +44,6 @@ The ball has to be placed at the position from where the ball was kicked (see th
 After the ball has been placed, a <<Free Kick, free kick>> is awarded to the opponent of the team that last touched the ball before it left the field.
 
 .Usage
-A kick is aimless when after the ball touched a robot, it subsequently crossed the midline and then its opponent's goal line outside the goal without touching another robot.
+A kick is aimless when after the ball touched a robot, it subsequently crossed the halfline and then its opponent's goal line outside the goal without touching another robot.
 
 A kick-off kick cannot be aimless, as the ball is located at the midline and does therefore not cross it.
```
kkimurak, line 49: Sorry for late but this "midline" also be affected?

g3force: indeed. I want to update the definition of those lines anyway, so that they will also have references. I'll probably do this in the upcoming weeks.

kkimurak: Alright, thanks.

PR #54, opened 2022-10-09
More explicit field coloring

g3force:
- Define the boundary walls to be black
- Define the goals to be black outside and white inside


masa-ito4github (approved):
LGTM!

PR #62, opened 2022-11-20
Add guidance for referees on unsporting behavior

g3force:


chapters/offenses.adoc
```diff
@@ -29,6 +34,13 @@ It is not allowed to damage or modify robots of other teams.
 ==== Damaging The Field Or The Ball
 It is not allowed to damage or modify the field or the ball.
 
+==== Disrespect Procedures
+Not following defined procedures repetitively, like for example:
+
+* <<Robot Handler, Robot handler>> puts a robot on the field, while it is not allowed
+* Robots do not keep required distance to the ball during stop
+* Keeper is not on the goal line during a <<Penalty Kick, penalty kick>>
```
lhcavalcanti: Can we add here:  Manually removing misplaced robots in Penalty Kick ?

g3force: How about a more general the statement:
```suggestion
* Robots do not conform to the positioning rules during a <<Penalty Kick, penalty kick>> and need to be moved or removed manually
```

PR #65, opened 2022-12-18
Update consequence for not removing robots on yellow cards

g3force:


tobiasheineken (commented):
I would prefer to _just_ remove the 2023 braced expression `(probably 2023)` Removing the idea of what we are probably going to introduce disallows planning.

g3force:
> I would prefer to _just_ remove the 2023 braced expression `(probably 2023)` Removing the idea of what we are probably going to introduce disallows planning.

I thought this is rather misleading, as it is not clear at all if this will be the penalty. Actually, we drafted something completely different recently, before we postponed it again...

Do you prefer that there is _some_ possible penalty, rather than none?

tobiasheineken (approved):

PR #71, opened 2024-01-20
Limiting group phase DivA robots to 8

lhcavalcanti:
As proposed during RoboCup 2023 and to help teams move from division B to division A. During the division A group phase, matches will be played with a maximum of 8 robots. In the Division A knockout phase, teams can play with up to 11 robots.

### DIFF
--- a/chapters/robots.adoc
+++ b/chapters/robots.adoc
@@ -3,7 +3,7 @@
 === Number Of Robots
 A match is played by two teams, with each team consisting of not more than 11 robots in division A and not more than 6 robots in division B, one of which may be the keeper. Each robot must be clearly numbered according to its vision pattern so that the referee can identify it during the match. The id of the keeper must be chosen before the match starts (see <<Choosing Keeper Id>>).
 
-To incentivize teams to move from division B to division A, during the division A group phase, matches will be played with a maximum of 8 robots. If both teams have more than 8 robots, they can play with up to 11 robots. In the division A knockout phase, teams can play with up to 11 robots.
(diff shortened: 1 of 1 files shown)


chapters/robots.adoc
```diff
@@ -3,6 +3,8 @@
 === Number Of Robots
 A match is played by two teams, with each team consisting of not more than 11 robots in division A and not more than 6 robots in division B, one of which may be the keeper. Each robot must be clearly numbered according to its vision pattern so that the referee can identify it during the match. The id of the keeper must be chosen before the match starts (see <<Choosing Keeper Id>>).
 
+To incentivize teams to move from division B to division A, during the division A group phase, matches will be played with a maximum of 8 robots. If both teams have more than 8 robots, they can play with up to 11 robots. In the division A knockout phase, teams can play with up to 11 robots.
```
g3force: I would remove the explanation/motivation part from this sentence. You can add it in an INFO box, but for the rules itself, the reasoning is not relevant.

Try to keep the sentence short and precise. And avoid repeating previously written rules.

> In the division A knockout phase, teams can play with up to 11 robots.

This sounds like teams can choose any max number of bots between 8 and 11. This is not, what we intended, right?

I'd try to formulate the rule along the lines of "An exception is made during the group phase in division A, where the number of robots is limited to 8, if at least one of the two teams prefers that."

lhcavalcanti: Got it. Indeed, it is way simpler and clearer.

Thanks for the suggestion, just made the fix.

g3force (approved):

PR #78, opened 2024-02-15
Add a punishment for robot substitution

### LINKED ISSUE #75: Robot substitution - Punishment
# Targets
- Reduce the used match time needed to substitute robots
- Teams should be encouraged to substitute quick and only do substitution
- Teams should be encouraged to make their robots more reliable and use automatic substitution, if possible. 

# Constraints
- Teams always need to be able to substitute robots. There can not be a hard limit
- It must be safe for robot handlers - teams should not be encouraged to move as fast as possible
- Automatic substitution during a match is always preferred

# Proposed Change
- Give teams 5 free substitutions per halftime
  - In FIFA, [five substitutions are allowed](https://www.fifa.com/fifaplus/en/member-associations/france/articles/all-you-need-to-know-about-fifa-world-cup-qatar-2022-squad-lists)
- Each substitution gives the team a time budget of 10s to substitute as many robots as they like. If they do not stop the substitution in time, a new substitution is started.
- Introduce a new foul game event that is raised, when a team starts a substitution and the free budget is used up.
  - The foul increases the foul counter
  - The foul results in a free kick for the opponent team
  - If both teams perform robot substitution in the same stoppage, the original command is issued instead of the freekick
- We will evaluate the numbers in upcoming regional competitions and may adapt them for RoboCup

g3force:
closes #75

### DIFF
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -176,3 +176,8 @@ NOTE: This rule does not cover all cases of ball placement interference.
 The <<Referee, referee>> is encouraged to call fouls if the non-placing team is obviously interfering with the ball placement.
 
 NOTE: If a robot keeps interfering the ball placement (for example if it is stuck or can not move), the human referee is encouraged to stop the placement and place the ball manually.
+
+===== Excessive Robot Substitutions
+If a team has used up their free robot substitution budget, every additional robot substitution is a foul.
+The match is resumed with a <<Free Kick, free kick>> for the opponent team.
+If both teams committed this foul in the same <<Stop, stop>>, the match is resumed with the original command.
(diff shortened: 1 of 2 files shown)


g3force:
Observation: If the GC issues an automatic robot substitution due to too many robots, the team is not prepared and 10s to perform the substitution gets tough.

g3force:
The automatic robot substitution game event, that is currently issued by the GC, should be removed. The game event is only triggered when the team actively requests the substitution through the remote-control or (as a fallback) through GCO. This is a GC implementation detail and not part of the rules. It currently automatically requests robot substitution on TOO_MANY_ROBOTS game event.

PR #86, opened 2024-12-13
Robot's Hull Color

lhcavalcanti:
The purpose of this PR is to integrate 2024's Technical Challenge as part of official games in RoboCuop 2025 for both divisions (A & B).

2024 TC rules: https://robocup-ssl.github.io/technical-challenge-rules/2024-ssl-black-white-hulls-rules.pdf

### DIFF
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -97,7 +97,7 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
 ==== Choosing Robot Hull Colors
-The <<Hull Color of Robots, robot hull color>> (dark or bright) should match the <<Choosing Team Colors, team's color>>. The blue team uses the dark hulls, while the yellow team uses light hulls.. If teams switch color during the game, they must switch also hull colors.
+The <<Hull Color of Robots, robot hull color>> (dark or bright) should match the <<Choosing Team Colors, team's color>>. The blue team uses the dark hulls, while the yellow team uses light hulls. If teams switch color during the game, they must switch also hull colors.
 
 ==== Choosing Side And Kick-Off
 The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers>>. The winning team chooses the goal it will attack in the first half of the match. The other team takes the <<Kick-Off, kick-off>> to start the match.
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -97,7 +97,7 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
 ==== Choosing Hulls Colors
-The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> for their preferred hull color (dark or bright). If teams agree, the colors are used; otherwise, the referee assigns them by chance.
+The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> for their preferred <<Robot's Hull Color, hull color>> (dark or bright). If teams agree, the colors are used; otherwise, the referee assigns them by chance.
 
 Robots must keep their assigned color throughout the game.
 
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -96,8 +96,8 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
-==== Choosing Hulls Colors
-The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> for their preferred <<Robot's Hull Color, hull color>> (dark or bright). If teams agree, the colors are used; otherwise, the referee assigns them by chance.
+==== Choosing Robot Hull Colors
+The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> for their preferred <<Hull Color of Robots, robot hull color>> (dark or bright). If teams agree, the colors are used; otherwise, the referee assigns them by chance.
 
 Robots must keep their assigned color throughout the game.
 
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -96,6 +96,11 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
+==== Choosing Hulls Colors
+The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> for their preferred hull color (dark or bright). If teams agree, the colors are used; otherwise, the referee assigns them by chance.
+
+Robots must keep their assigned color throughout the game.
+
 ==== Choosing Side And Kick-Off
 The <<Referee, referee>> tosses a coin with both <<Robot Handler, robot handlers>>. The winning team chooses the goal it will attack in the first half of the match. The other team takes the <<Kick-Off, kick-off>> to start the match.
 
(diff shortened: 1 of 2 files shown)


chapters/robots.adoc
```diff
@@ -54,16 +67,3 @@ NOTE: Bluetooth is not allowed since it cannot be fixed to frequency channels.
 
 ==== Autonomy
 The robotic equipment has to be fully autonomous. Human operators are not permitted to enter any information to the system during a match, except in <<Overview, breaks>> or during a <<Timeouts,timeout>>. Disregarding this rule is considered <<Unsporting Behavior, unsporting behavior>>.
-.Standard Vision Pattern Colors
```
g3force: Why is this removed?

chapters/gamestructure.adoc
```diff
@@ -166,8 +171,3 @@ the team with more goals is declared the winner.
 
 NOTE: During the group phase, the number of goals scored is used as tie-breaker, so the absolute number of goals matter for overall scoring.
 The rule applies to all game types for simplicity.
-the match is automatically terminated as soon as the goal difference is greater than one and
```
g3force: Why is this removed?

chapters/gamestructure.adoc
```diff
@@ -96,6 +96,11 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
+==== Choosing Hulls Colors
+The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> for their preferred hull color (dark or bright). If teams agree, the colors are used; otherwise, the referee assigns them by chance.
```
g3force: A reference to the other section would be nice here.

chapters/robots.adoc
```diff
@@ -18,6 +18,19 @@ The <<Referee, referee>> has to force a team to remove a robot from the field if
 ==== Shape
 A robot must fit inside a 0.18 meters wide and 0.15 meters high cylinder at any point in time. Additionally, the top of the robot must adhere to the standard pattern size and surface constraints.
 
+==== Robot's Hull Color
+To ensure clarity and effective identification during gameplay, the following rules apply to the coloring of robot hulls:
+
+Bright and Dark Color Scheme:
+* Each robot must include both a bright and a dark color in its design.
```
g3force: This could also be interpreted as: Robot must have both colors at the same time.

chapters/robots.adoc
lhcavalcanti: It was by mistake. Fixed

chapters/gamestructure.adoc
lhcavalcanti: It was by mistake. Fixed

chapters/gamestructure.adoc
lhcavalcanti: added!

chapters/robots.adoc
lhcavalcanti: Made some changes. Could you take another look?

chapters/gamestructure.adoc
```diff
@@ -96,6 +96,11 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
+==== Choosing Hulls Colors
```
g3force: Shouldn't it be "Choosing Hull Colors"? Or more explicit "Choosing Robot Hull Colors". Or "Choosing Color of the Robot Hulls".

chapters/robots.adoc
```diff
@@ -18,6 +18,22 @@ The <<Referee, referee>> has to force a team to remove a robot from the field if
 ==== Shape
 A robot must fit inside a 0.18 meters wide and 0.15 meters high cylinder at any point in time. Additionally, the top of the robot must adhere to the standard pattern size and surface constraints.
 
+==== Robot's Hull Color
```
g3force: Not totally sure, but I think `'` is wrong here. Maybe rewrite it to "Hull Color of Robots".

chapters/robots.adoc
```diff
@@ -18,6 +18,22 @@ The <<Referee, referee>> has to force a team to remove a robot from the field if
...
+To ensure clarity and effective identification during gameplay, the following rules apply to the coloring of robot hulls:
+
+Bright and Dark Color Scheme:
```
g3force: As this is not a title, it should be all lowercase, except for first word.

chapters/robots.adoc
```diff
@@ -18,6 +18,22 @@ The <<Referee, referee>> has to force a team to remove a robot from the field if
...
+
+* All robots from a team must include two interchangeable colors: one bright and one dark.
```
g3force: Hmm. They do not have to (physically) include them, right?

chapters/robots.adoc
```diff
@@ -18,6 +18,22 @@ The <<Referee, referee>> has to force a team to remove a robot from the field if
...
+* These two colors must be the same for all robots in a team.
+* The color must cover at least 40% of the robot's maximum height (15 cm), equivalent to a minimum of 6 cm of vertical coverage.
```
g3force: What, if the robot is smaller? Is the actual robot height, or the allowed robot height considered?

chapters/robots.adoc
```diff
@@ -18,6 +18,22 @@ The <<Referee, referee>> has to force a team to remove a robot from the field if
...
+* Teams must use the same color scheme throughout the entire game.
```
g3force: Isn't this redundant and could be skipped here?

chapters/gamestructure.adoc
lhcavalcanti: Yep, I changed for the explicit one.

chapters/robots.adoc
lhcavalcanti: I was debating with myself on this. It could be 40% of the robot's actual height. I'm afraid someone comes with a small robot, so the 40% will not be visible. But I don't think we will have robots lower than 10 cm. 😅

chapters/robots.adoc
lhcavalcanti: Good catch; in my mind, I was not considering "include" as having both of them physically. Reworded it, let me know if now sounds better.

chapters/robots.adoc
g3force: I'd make it fixed length for simplicity. That's also easier to check ;)

lhcavalcanti: Sure! I will fix in 6cm, I think it will be enough to view the robot's difference.

chapters/gamestructure.adoc
```diff
@@ -96,6 +96,11 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
+==== Choosing Robot Hull Colors
+The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> for their preferred <<Hull Color of Robots, robot hull color>> (dark or bright). If teams agree, the colors are used; otherwise, the referee assigns them by chance.
+
+Robots must keep their assigned color throughout the game.
```
tobiasheineken: So we disconnect the real team colors (blue / yellow) that can switch in half time and these hull colors that cannot? Is that intentional, or just an oversight?

I was thinking about simply always connecting the "bright" color with yellow, and the "dark" with blue, which would make this very easy on the referee and the status board. But then we would have teams change their hulls during half time, if the pattern colors are contested.

lhcavalcanti: I was unsure about that. I agree that it simplifies things. My only concern is to cause a lot of hull color changes in the intervals. I will update with this connection here.

chapters/gamestructure.adoc
```diff
@@ -96,6 +96,9 @@ The <<Referee, referee>> asks the <<Robot Handler, robot handlers>> of the teams
 
 However, if both teams prefer the same color, the referee assigns the colors by chance. In this case, the teams switch the colors after the first half of the match as well as after the first half of the overtime if applicable.
 
+==== Choosing Robot Hull Colors
+The <<Hull Color of Robots, robot hull color>> (dark or bright) should match the <<Choosing Team Colors, team's color>>. The blue team uses the dark hulls, while the yellow team uses light hulls.. If teams switch color during the game, they must switch also hull colors.
```
```suggestion
The <<Hull Color of Robots, robot hull color>> (dark or bright) should match the <<Choosing Team Colors, team's color>>. The blue team uses the dark hulls, while the yellow team uses light hulls. If teams switch color during the game, they must switch also hull colors.
```

PR #88, opened 2025-01-07
Added rule to 4.4.2 Timeouts

### LINKED ISSUE #87: Prevent teams from manually moving non-functional robots in Timeout phase
During the RoboCup 2024, teams used the Timeout phase to manually move non-functioning robots to more favourable positions on the field. This is not in the spirit of the Small Size League, and rule(s) should be implemented to prevent this kind of gameplay

emielsteerneman:
This draft PR adds the following to section 4.4.2 Timeouts:
```
Any robot that has been physically interacted with during a timeout must be treated as substituted, as outlined in <<Robot Substitution, Robot Substitution>>. Such a robot must be either removed from the field or placed in the area defined in <<Robot Substitution, Robot Substitution>> before the Timeout phase ends.
```

### DIFF
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -150,7 +150,7 @@ NOTE: As a result, the time needed for a match is much greater than the playing 
 ==== Timeouts
 The <<Robot Handler, robot handler>> has to ask the referee for a timeout. Timeouts are handled like <<Overview, breaks>>, meaning that both teams are allowed to make modifications to their software and hardware (see <<Autonomy>>).
 
-Any robot that has been physically interacted with during a timeout must be treated as substituted, as outlined in <<Robot Substitution, Robot Substitution>>. Such a robot must be either removed from the field or placed in the area defined in <<Robot Substitution, Robot Substitution>> before the Timeout phase ends.
+Any robot that has been physically interacted with during a timeout must be taken out of the field. It can be brought in again, also during the same timeout, via the area outlined in <<Robot Substitution, Robot Substitution>>.
 
 Each team is allocated 4 timeouts at the beginning of the match. A total of 300 seconds is allowed for all timeouts. Timeouts may only be taken during a game
 stoppage. The time is monitored and recorded by the <<Game Controller Operator, game controller operator>>.
--- a/chapters/gamestructure.adoc
+++ b/chapters/gamestructure.adoc
@@ -150,7 +150,7 @@ NOTE: As a result, the time needed for a match is much greater than the playing 
 ==== Timeouts
 The <<Robot Handler, robot handler>> has to ask the referee for a timeout. Timeouts are handled like <<Overview, breaks>>, meaning that both teams are allowed to make modifications to their software and hardware (see <<Autonomy>>).
 
-Any robot that has been physically interacted with during a timeout must be treated as substituted, as outlined in <<Robot Substitution, Robot Substitution>>. Such a robot must be either removed from the field or placed in the area defined in <<Robot Substitution, Robot Substitution>> before the Timeout phase ends.
+Any robot that has been physically interacted with during a timeout must be treated as substituted, as outlined in <<Robot Substitution, Robot Substitution>>. Such a robot must be either removed from the field or placed in the area defined in <<Robot Substitution, Robot Substitution>> before the timeout phase ends.
 
 Each team is allocated 4 timeouts at the beginning of the match. A total of 300 seconds is allowed for all timeouts. Timeouts may only be taken during a game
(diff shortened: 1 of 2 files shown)


chapters/gamestructure.adoc
```diff
@@ -150,6 +150,8 @@ NOTE: As a result, the time needed for a match is much greater than the playing
 ==== Timeouts
 The <<Robot Handler, robot handler>> has to ask the referee for a timeout. Timeouts are handled like <<Overview, breaks>>, meaning that both teams are allowed to make modifications to their software and hardware (see <<Autonomy>>).
 
+Any robot that has been physically interacted with during a timeout must be treated as substituted, as outlined in <<Robot Substitution, Robot Substitution>>. Such a robot must be either removed from the field or placed in the area defined in <<Robot Substitution, Robot Substitution>> before the Timeout phase ends.
```
```suggestion
Any robot that has been physically interacted with during a timeout must be treated as substituted, as outlined in <<Robot Substitution, Robot Substitution>>. Such a robot must be either removed from the field or placed in the area defined in <<Robot Substitution, Robot Substitution>> before the timeout phase ends.
```

tobiasheineken:
I like the change. Do we also want a change to 3.2.6 to make clear moving a robot by hand is not acceptable during a match?

emielsteerneman:
> I like the change. Do we also want a change to 3.2.6 to make clear moving a robot by hand is not acceptable during a match?

At the risk of being pedantic:
1. Does moving robots into the field during the time before the start signal or break fall under "match"?
2. Moving by hand is acceptable when moving robots out of the field, though I get your point


Additionally, maybe I should reword `substituted` in the rule change with `taken out of the field`. Marking a robot as substituted also implies that during timeout, robots can only be taken out of the field by a robot handler. It may also be confusing with the rule `Each team has 5 free substitutions per halftime`. Does a robot marked for substitution during timeout count towards this? I'd say no, but it's still confusing.

Edit: Or does it actually mean `Each team has 5 free substitution INTENTS per halftime`, not counting when a robot drives itself to the exchange area to be picked up while the game continues?

@tobiasheineken @g3force

g3force:
I'd say, a substitution INTENT is free. You can revoke it without loosing anything:

> A substitution intent can be revoked unless the game was not already halted for substitution.

Once the intent is accepted by the GC, its converted into a substitution (which allows for arbitrary many robots to substitute). This counts towards your substitution budget.

g3force:
I'd also change "substituted" to sth like: 

Any robot that has been physically interacted with during a timeout must be taken out of the field. It can be brought in again as outlined in <<Robot Substitution, Robot Substitution>>.

g3force:
> Do we also want a change to 3.2.6 to make clear moving a robot by hand is not acceptable during a match?

Probably worth adding.
"The robotic equipment has to be fully autonomous" also has some room for interpretation. One could argue, that this means that all robots must be able to move (what we used to imply so far), but that's quite implicit.

> Does moving robots into the field during the time before the start signal or break fall under "match"?

I'd say that's not match.
However, if we would want to be really strict, we could ask teams to move all robots to the substitution area before the match, so that the robots are forced to move.
That would be easier, if we'd had an explicit substitution area, though. Because than we would not even care, if the robots would not move. We could consider this next year, when we evaluate such an area.

emielsteerneman:
I reworded `4.4.2 Timeouts` and added to `3.2.6 Autonomy`

chapters/robots.adoc
```diff
@@ -53,4 +53,4 @@ The type of wireless communication may also be restricted by the <<Local Organiz
 NOTE: Bluetooth is not allowed since it cannot be fixed to frequency channels.
 
 ==== Autonomy
-The robotic equipment has to be fully autonomous. Human operators are not permitted to enter any information to the system during a match, except in <<Overview, breaks>> or during a <<Timeouts,timeout>>. Disregarding this rule is considered <<Unsporting Behavior, unsporting behavior>>.
\ No newline at end of file
+The robotic equipment has to be fully autonomous. Human operators are not permitted to enter any information to the system during a match, except in <<Overview, breaks>> or during a <<Timeouts,timeout>>. Additionally, robots may not be manually moved during any phase of the match, except for moving robots out of the field, or into the field as described in <<Robot Substitution, Robot Substitution>>. Disregarding this rule is considered <<Unsporting Behavior, unsporting behavior>>.
```
g3force, line 56: Small tip: It is good practice to put each sentence on a separate line. That's easier to read and makes it easier to compare the GIT diff.

tobiasheineken:
Looks good to me, thank you!

g3force (approved):

PR #104, opened 2026-05-09
Rename Emergency Stop to Force Timeout

guyfleeman:
This sets the stage to rename "Emergency Stop" in the league rules and league software. The league discussed that the name "Emergency Stop" generally doesn't describe what the action does and has negative safety implications. 

The event has a 10 second delay and is intended for situations where robots have lost partially lost control or software crashes. These are not emergencies. Neither of these are resolved by a field level e-stop and would still require physical intervention by the team either with the robot or the laptop. Also, the 10 second delay is the exact opposite of the expected behavior of "emergency stop" which is that true emergencies should be addressed without delay.

The action is now named "Force Timeout" (verb) and produces a "Forced Timeout" (noun) game state.

Clarify rules under the "Halt" section, that unsafe behavior by robots (e.g. loss of control, fire, electronics failures like unintended arc emission from kickers) is handled at the discretion of the referee, who may immediately halt the game at any time. The referee may penalize teams for repeated robot behavior that makes him question or conclude safety has been compromised.

### DIFF
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -34,6 +34,9 @@ It is not allowed to damage or modify robots of other teams.
 ==== Damaging The Field Or The Ball
 It is not allowed to damage or modify the field or the ball.
 
+==== Unsafe Operations
+It is not allowed to have robot failures that endanger the participants or scpectators. Examples of unsafe operations include, but not limited to: high power or high velocity uncontrolled robot operation, electronics fire, high voltage or visible arc discharge.
+
 ==== Disrespect Procedures
 Not following defined procedures repetitively, like for example:
 
--- a/chapters/offenses.adoc
+++ b/chapters/offenses.adoc
@@ -35,7 +35,7 @@ It is not allowed to damage or modify robots of other teams.
 It is not allowed to damage or modify the field or the ball.
 
 ==== Unsafe Operations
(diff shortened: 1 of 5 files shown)


chapters/offenses.adoc
```diff
@@ -34,6 +34,9 @@ It is not allowed to damage or modify robots of other teams.
 ==== Damaging The Field Or The Ball
 It is not allowed to damage or modify the field or the ball.
 
+==== Unsafe Operations
+It is not allowed to have robot failures that endanger the participants or scpectators. Examples of unsafe operations include, but not limited to: high power or high velocity uncontrolled robot operation, electronics fire, high voltage or visible arc discharge.
```
```suggestion
It is not allowed to have robot failures that endanger the participants or spectators. Examples of unsafe operations include, but not limited to: high power or high velocity uncontrolled robot operation, electronics fire, high voltage or visible arc discharge.
```

what is meant by "high power"? The last sentence seems a bit rough.

nerdfire01: I'd guess, that high power could cover situations, where the wheels of a robot turn in opposite directions, such that the robot doesn't actually move but strips away carpet.
Therefore I'd be fine with it under the goal, to get the updated rules in a timely manner.

g3force (approved):

